Issue #16, opened 2022-11-25
Golang by @Reanon


krahets:
Translating codes from Java to Golang.

PR #20, opened 2022-11-25
feat(go): support basic pkg

Reanon:
1. add list_node & tree_node

### DIFF
--- a/codes/go/pkg/list_node.go
+++ b/codes/go/pkg/list_node.go
@@ -1,0 +1,59 @@
+// File: list_node.go
+// Created Time: 2022-11-25
+// Author: Reanon (793584285@qq.com)
+
+package pkg
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+// ListNode Definition for a singly-linked list node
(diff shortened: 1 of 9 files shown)


codes/go/pkg/list_node.go
```diff
@@ -0,0 +1,59 @@
...
+// NewListNode Generate a list node with an val
+func NewListNode(v int) *ListNode {
+	return &ListNode{
+		Next: nil,
+		Val:  v,
+	}
+}
+
+// ArrayToLinkedListLinkedList Generate a linked list with an array
+func ArrayToLinkedListLinkedList(arr []int) *ListNode {
```
krahets, line 28: Is this `ArrayToLinkedList` ?

krahets (commented):
Thanks for the great work! There is a comment that need to be addressed, thanks!

codes/go/pkg/list_node.go
Reanon: Amended in next commit.

krahets (approved):
Thanks! Now merge it.
#16

PR #29, opened 2022-11-26
Support go code in chapter-tree

Reanon:


codes/go/pkg/tree_node.go
```diff
@@ -9,6 +9,9 @@ import (
 	"fmt"
 )
 
+// nodeNotExist represents node don't exist.
+const nodeNotExist int = -1
```
krahets, line 13: Does this definition work if we have a node with val = -1?

codes/go/chapter_tree/binary_search_tree_test.go
```diff
@@ -0,0 +1,44 @@
+// File: binary_search_tree_test.go
+// Created Time: 2022-11-26
+// Author: Reanon (793584285@qq.com)
+
+package chapter_tree
+
+import "testing"
+
+func TestBinarySearchTree(t *testing.T) {
+	nums := []int{8, 9, 10, 11, 12, 13, 14, 15, 1, 2, 3, 4, 5, 6, 7}
```
krahets, line 10: It is not consistent with the Java version, i.e. `int[] nums = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15 };`
Did you use a different implementation?

codes/go/chapter_tree/binary_search_tree_test.go
```diff
@@ -0,0 +1,44 @@
...
+	// 插入结点
+	node = bst.Insert(16)
+	t.Log("插入结点后 16 的二叉树为: ")
+	bst.Print()
+
+	// 删除结点
+	bst.Remove(1)
+	t.Log("删除结点 1 后的二叉树为: ")
+	bst.Print()
+	bst.Remove(12)
```
krahets, line 35: Recommend removing it for better consistency among multiple languages.
(Cause maybe someone will translate from Go to another language, which has a chain reaction.

krahets (commented):
Thanks for the clean code!

codes/go/pkg/tree_node.go
Reanon: I see. I have find a better way to solve this.

codes/go/chapter_tree/binary_search_tree_test.go
Reanon: Fixed in next commit.

codes/go/chapter_tree/binary_search_tree_test.go
Reanon: Fixed in next commit.

krahets (approved):
Thanks, @Reanon !

PR #50, opened 2022-11-29
栈的python代码

pengchzn:


codes/python/chapter_stack_and_queue/array_stack.py
```diff
@@ -4,7 +4,78 @@
 Author: Krahets (krahets@163.com)
```
krahets, line 4: Please update the author and time information in the file header with yours, thanks!

codes/python/chapter_stack_and_queue/array_stack.py
```diff
@@ -4,7 +4,78 @@
...
+""" 基于数组实现的栈 """
+
+
+class ArrayStack:
+    def __init__(self):
+        self._stack = []
+        self._size = 0
+
+    """ 获取栈的长度 """
+
```
krahets, line 22: No empty line between the function name and the function

codes/python/chapter_stack_and_queue/array_stack.py
```diff
@@ -4,7 +4,78 @@
...
 
-import sys, os.path as osp
+import os.path as osp
+import sys
+
 sys.path.append(osp.dirname(osp.dirname(osp.abspath(__file__))))
 from include import *
 
+""" 基于数组实现的栈 """
+
```
krahets, line 14: No empty line here

krahets (commented):
Thanks, @pengchzn ! Some comments need to be addressed.

pengchzn:
Got it, I've already corrected them, thanks!

codes/python/chapter_stack_and_queue/array_stack.py
```diff
@@ -1,10 +1,73 @@
...
+    pop = stack.pop()
+    print("出栈元素 pop = ", pop)
+    print("出栈后 stack = ", stack._stack)
+
+    """ 获取栈的长度 """
+    size = stack.size()
+    print("栈的长度 size = ", size)
+
+    """ 判断是否为空 """
+    isEmpty = (stack == [])
```
krahets, line 73: This should be `isEmpty = stack.isEmpty()`.

And `stack` is a ArrayStack object, which will never satisfy `stack == []`

codes/python/chapter_stack_and_queue/array_stack.py
```diff
@@ -1,10 +1,73 @@
...
+    def __init__(self):
+        self._stack = []
+        self._size = 0
+
+    """ 获取栈的长度 """
+    def size(self):
+        return self._size
+
+    """ 判断栈是否为空 """
+    def isEmpty(self):
```
krahets, line 24: is_emtpy()

codes/python/chapter_stack_and_queue/linkedlist_stack.py
```diff
@@ -1,10 +1,84 @@
...
+
+    """ 访问栈顶元素 """
+    def peek(self):
+        return self.head.val
+
+    def to_array(self):
+        stack = []
+        temp = self.head
+        while temp:
+            stack.insert(0, temp.val)
```
krahets, line 53: It is recommended `append(temp.val)` here, and followed by  `temp.reverse()` to reduce time complexity.

codes/python/chapter_stack_and_queue/linkedlist_stack.py
```diff
@@ -1,10 +1,84 @@
...
+    pop = stack.pop()
+    print("出栈元素 pop = ", pop)
+    print("出栈后 stack = ", stack.to_array())
+
+    """ 获取栈的长度 """
+    size = stack.size()
+    print("栈的长度 size = ", size)
+
+    """ 判断是否为空 """
+    isEmpty = (stack == [])
```
krahets, line 84: stack will never satisfy `stack == []` You should use `stack.is_empty()`

krahets (changes requested):
A few more comments, thanks!

codes/python/chapter_stack_and_queue/linkedlist_stack.py
pengchzn: The ListNode type seems doed not have the function of reverse(), so I slice the index of it instead, that should be faster than using reverse

pengchzn: Got it, I slice the index of it instead, for it should be faster than using reverse(),thanks for comments

krahets (approved):

krahets:
Thanks for the code! @pengchzn 

By the way, `reverse()` is better than slicing to reverse a list. More information at

https://www.geeksforgeeks.org/python-reversed-vs-1-which-one-is-faster/

pengchzn:
Ah, I will check that article for more information, maybe I keep the wrong idea for quite a long time. Thanks for that

krahets:
> Ah, I will check that article for more information, maybe I keep the wrong idea for quite a long time. Thanks for that

https://github.com/krahets/hello-algo/commit/53cc651af2ac03f69f7b68d21fbecbbda747cf85

I update the implementation of the array stack, especially the linked list stack. Please check it.

PR #53, opened 2022-11-30
Add python code for queue

pengchzn:


codes/python/chapter_stack_and_queue/array_queue.py
```diff
@@ -1,10 +1,95 @@
...
+        return len(self.nums)
+
+    """ 获取队列的长度 """
+    def size(self):
+        # 由于将数组看作为环形，可能 rear < front ，因此需要取余数
+        return (self.capacity() + self.rear - self.front) % self.capacity()
+
+    """ 判断队列是否为空 """
+    def is_empty(self):
+        return len(self.nums)==0
```
krahets, line 31: `return rear - front == 0` . Please check your code compared with Java/C++ code, thanks!

krahets (changes requested):
Thanks, @pengchzn ! There are comments need to be addressed.

krahets (approved):
Thank @pengchzn for the hard work!

PR #55, opened 2022-11-30
Update array_stack.py

pengchzn:
An type error of ";"


krahets (approved):
Thank you for your attentiveness!

PR #56, opened 2022-12-01
fix(codes/cpp/include/TreeNode.hpp):  fix 'vecToTree' function

mgisr:


krahets (approved):
Thanks for the correction!

PR #95, opened 2022-12-12
Add the remain JavaScript code and docs  and Update the TypeScript style for Chapter of Array and Linkedlist

justin-tse:


krahets:
LGTM! I have to clarify that due to conflicting codes with another PR, I merge them and make the two contributors "coauthor" the codes.

PR #99, opened 2022-12-13
time complexity using go

msk397:

PR #102, opened 2022-12-13
feat(time-complexity): code implemented in Go

Cathay-Chen:
add go code for chapter_computational_complexity/time_complexity.


krahets:
Thank @Cathay-Chen for the great work! There is a previous pull request #99 , so I merged your work and made the file co-authored by the two authors.

By the way, two tips for the code:

1. Use `go-format` to format the code.
2. Prefer `fmt.Print()` rather than `Print()`.

Best regards

PR #104, opened 2022-12-13
Fix the PrintUtil.js showTrunks and Update JavaScript and docs' style (Chapter of Tree)

justin-tse:
PrintUtil.js:
console.log(p.str) will print a new empty line, which will affect the tree print.
image
so we should change to process.stdout.write(p.str) because it will not print an empty line.
image


krahets (approved):
LGTM. Thank you for the code review and bug fixing!

PR #114, opened 2022-12-15
Update intro_to_sort.md

liuxjerry:
错别字修正


krahets (approved):
Thanks for the correction!

PR #125, opened 2022-12-17
Add C# linkedlist doc and code,modify some formatting.

mingXta:
- Add C# linkedlist doc and code.
- Modify some formatting to be more in line with the C# coding specification,lile:
 `/* 随机返回一个数组元素 */`-> `/// <summary>
        /// 随机返回一个数组元素
       /// </summary>`
- Create a project to replace single files,and add unit test.(This way others can run it directly without any other operations)


codes/csharp/chapter_array_and_linkedlist/Array.cs
```diff
@@ -1,14 +1,14 @@
...
+// File: Array.cs
+// Created Time: 2022-12-14
+// Author: mingXta (1195669834@qq.com)
+
 namespace hello_algo.chapter_array_and_linkedlist
 {
     public class Array
     {
-        /* 随机返回一个数组元素 */
+        /// <summary>
```
krahets, line 9: Thank you for updating the function headers to the C# standard format. However, it is recommended to use the `/* ... */` format to make the codes more compact because the codes are also shown in markdown files.

codes/csharp/hello-algo.sln
```diff
@@ -0,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.1.32421.90
```
krahets, line 4: Could we use a more lightweight IDE to run the code, like VSCode? And is it possible to make every source code file independent? That is, the users could copy just one file and run it, not the whole directory.

codes/csharp/test/chapter_array_and_linkedlist/ArrayTest.cs
```diff
@@ -0,0 +1,70 @@
+﻿// File: ArrayTest.cs
+// Created Time: 2022-12-16
+// Author: mingXta (1195669834@qq.com)
+
+using NUnit.Framework;
```
krahets, line 5: Could we merge `ArrayTest.cs` and `Array.cs` . Compared to the engineering correctness, the simplicity is more significant for this book.

krahets (commented):
Thanks for the clean codes! Please address the comments.

codes/csharp/hello-algo.sln
mingXta: I will delete this file and the project can be run in every editor or ide.But for C#, running a single file is unusual, and for systems other than Win that must rely on other tools.

codes/csharp/test/chapter_array_and_linkedlist/ArrayTest.cs
mingXta: Sure, I'll modify it.

codes/csharp/chapter_array_and_linkedlist/Array.cs
mingXta: `///` can will give better ide support and I will use `/* ... */` in the md file,Is this okay？

codes/csharp/hello-algo.sln
krahets: That's OK. I‘m not familiar with C# (I regard Java as an alternative to C#, but I forget the multi-platform capability of the two languages has a large gap.)

Do you know a solution to make the C# code run on all common systems like Win, macOS, and ubuntu(Linux)? Cause the users could use different systems. Hope for your solution, thanks!

krahets: But for C#, running a single file is unusual

Hmmm, but I believe it is OK for learning leetcode and algorithm. Click one button running the code is the wish of us and this book.

codes/csharp/chapter_array_and_linkedlist/Array.cs
krahets: Yes, it is up to you~ However, in my opinion, it will make the work harder cause should modify comments for every function.

codes/csharp/hello-algo.sln
mingXta: The dotnet project supports multiple platforms, but compiling a single *.cs file can only be compiled into an exe.
When I found that I could not compile and run a single file on a mac which is why I created the project. The current project can run on any system with a .net environment.

I looked for information to find other systems can be installed https://www.mono-project.com/ to run *.cs files 
compiled exe, but I feel that this is a little weird.

C# is very similar to java, so it's very easy for me to translate java code (basically ctrl+c)

krahets: I see. Maybe this does be the best solution.

krahets (approved):
Thanks! Now merge it.

PR #126, opened 2022-12-17
function PrintSlice supports Generics

msk397:


msk397:
I found that Reanon completed the Go of  linear Search, but did not supplement the document, so I copied the code into the documentation

krahets (approved):
Thanks @msk397 for the great work!

PR #128, opened 2022-12-18
List using Go

msk397:


codes/go/chapter_array_and_linkedlist/list_test.go
```diff
@@ -0,0 +1,67 @@
...
+	/* 访问元素 */
+	num := list[1] // 访问索引 1 处的元素
+	fmt.Println("访问索引 1 处的元素，得到 num =", num)
+
+	/* 更新元素 */
+	list[1] = 0 // 将索引 1 处的元素更新为 0
+	fmt.Println("将索引 1 处的元素更新为 0 ，得到 list =", list)
+
+	/* 清空列表 */
+	list = []int{}
```
krahets, line 28: I think `list = []int{}` is to initialize but not clear a slice. It'd be better to use `list = nil` instead,

krahets (commented):
Thanks @msk397 ! Please address the comment.

codes/go/chapter_array_and_linkedlist/list_test.go
msk397: You are right, I have finished modifying!

krahets (approved):
Thanks for the work! Now merge it.

PR #154, opened 2022-12-21
Add array code in C

MolDuM:
My C code is mainly referred to C++ code since the similar grammar. The output of Java is also concerned.

There are two questions on the definition of array. 
1. malloc is not used to define an array in C++. But to free the space of the temporary array, I add an initializing way, but it is not smart.
2. Since malloc cannot initialize the array, I initialized it by myself. I am wondering if calloc should be introduced to reader in this part.

I hope the reviewer could give some advice. 

> Tip: If this PR is not related to the coding or code translation, please ignore the checklist.

### Checklist


codes/c/chapter_array_and_linkedlist/array.c
```diff
@@ -0,0 +1,124 @@
...
+    int randomIndex = rand() % size;
+    // 获取并返回随机元素
+    int randomNum = nums[randomIndex];
+    return randomNum;
+}
+
+/* 扩展数组长度 */
+int* extend(int* nums, int size, int enlarge) {
+    // 初始化一个扩展长度后的数组
+    int* res = (int *)malloc(sizeof(int) * (size + enlarge));
```
krahets, line 21: Sorry, I'm not familiar with C. But how about initializing the array in this way:

```c
int arr[size + enlarge];
int *res = arr;
```

codes/c/chapter_array_and_linkedlist/array.c
```diff
@@ -0,0 +1,124 @@
...
+    int size = 5;
+    int arr[5];
+    printf("数组 arr = ");
+    printArray(arr, size);
+
+    int arr2[5] = { 1, 3, 2, 5, 4 };
+    printf("数组 arr2 = ");
+    printArray(arr2, size);
+
+    int* nums = (int *)malloc(sizeof(int) * size);
```
krahets, line 83: Same as above.

krahets (commented):
Thank @MolDuM for the C codes! You are the first contributor writing in C. Please address the comments.

codes/c/chapter_array_and_linkedlist/array.c
MolDuM: I think it is not a good idea in C.
The first line defines a local variable array in the function, which means the address would be assigned to arr, and it would be freed when return. In another word, the numbers in arr would be lost after the function finished. Of course we can define a new array outside the function, but it would reduce the transportability.
Besides, the first line is not recommended in C because the length of array should be a constant value instead of a variable. There would be a reminder from IDE (but no warning and error) on my VSCode.
This is a common problem in C. In my experience, I would use malloc to define an unknown length array and use realloc to change the length. But it is not the classical way to define an array. I think I need to have a look at the official document and update it later.

krahets: Good explanation. 👍 

Actually, I think the array in C/C++ is too complex to use. In practice, We'd like to use `vector` as an array in C++ in 100% conditions.

MolDuM:
I have just added a commit. I finally chose a friendly way for beginner instead of the cleverest way.

reference:
[1] https://en.cppreference.com/w/c
[2] 《C程序设计》谭浩强

krahets (approved):
Thanks for the great work!

PR #167, opened 2022-12-26
docs(array.md): Added go array initialization code example

guowei-gong:


guowei-gong:
I think what is needed is a short statement, not multiple statemen.

PR #168, opened 2022-12-26
docs(array): Added golang array init code example

guowei-gong:

Code style I prefer the style of other codes in the book, instead of following the `go` programming convention.
e.g Single line comment not used `//`


guowei-gong:
Maybe I should merge the submission records, otherwise it looks strange.

PR #178, opened 2022-12-27
docs(array): sample code for golang

guowei-gong:
本次提交包含如下示例代码。
- 遍历数组；
- 初始化数组；
- 扩展数组长度；
- 在数组中查找指定元素；
- 随机返回一个数组元素；
- 删除索引 index 处元素；
- 在数组的索引 index 处插入元素 num。

✅ 已通过测试用例

所有数组约定长度为 5。原因如下：
在 goalng 中，必须声明数组的长度，例如：nums := [5]int{1,2,3,4,5}。如果不声明长度，则被称为切片。

使用的注释没有按照 golang 的编程惯例，而是倾向于使用文档上下文的注释约定。
所以所有函数注释均使用了 `/* ... */`，而不是双斜杠 `//`。

### DIFF
--- a/codes/go/chapter_array_and_linkedlist/array.go
+++ b/codes/go/chapter_array_and_linkedlist/array.go
@@ -5,23 +5,24 @@ import (
 	"math/rand"
 )
 
-// ExpectSize 预期大小。 在 Go 中，声明数组长度必须是常量表达式，
-// 所以这里我们约定，扩展后的长度为 6
-const ExpectSize = 6
+/**
+我们将 Go 中的 Slice 切片看作 Array 数组，降低理解成本，
+有利于我们将关注点放在数据结构与算法上。
+*/
 
 /* 随机返回一个数组元素 */
-func randomAccess(nums [5]int) (ans int) {
+func randomAccess(nums []int) (randomNum int) {
 	// 在区间 [0, nums.length) 中随机抽取一个数字
 	randomIndex := rand.Intn(len(nums))
 	// 获取并返回随机元素
-	ans = nums[randomIndex]
+	randomNum = nums[randomIndex]
 	return
 }
 
 /* 扩展数组长度 */
-func extend(nums [5]int) [ExpectSize]int {
+func extend(nums []int, enlarge int) []int {
 	// 初始化一个扩展长度后的数组
-	var res [ExpectSize]int
+	res := make([]int, len(nums)+enlarge)
 	// 将原数组中的所有元素复制到新数组
 	for i := 0; i < len(nums); i++ {
 		res[i] = nums[i]
@@ -31,7 +32,7 @@ func extend(nums [5]int) [ExpectSize]int {
 }
 
 /* 在数组的索引 index 处插入元素 num */
-func insert(nums *[5]int, num int, index int) {
+func insert(nums []int, num int, index int) {
 	// 把索引 index 以及之后的所有元素向后移动一位
 	// 如果超出了数组长度，会被直接舍弃
 	for i := len(nums) - 1; i > index; i-- {
@@ -42,11 +43,7 @@ func insert(nums *[5]int, num int, index int) {
 }
 
 /* 删除索引 index 处元素 */
-func remove(nums *[5]int, index int) {
-	// 越界检查
-	if index >= len(nums) {
-		return
-	}
+func remove(nums []int, index int) {
 	// 把索引 index 之后的所有元素向前移动一位
 	for i := index; i < len(nums); i++ {
 		if i+1 >= len(nums) {
@@ -58,7 +55,7 @@ func remove(nums *[5]int, index int) {
 }
 
 /* 遍历数组 */
-func traverse(nums [5]int) {
+func traverse(nums []int) {
 	var count int
 	// 通过索引遍历数组
 	for i := 0; i < len(nums); i++ {
@@ -71,11 +68,11 @@ func traverse(nums [5]int) {
 }
 
 /* 在数组中查找指定元素 */
-func find(nums [5]int, target int) (ans int) {
-	ans = -1
+func find(nums []int, target int) (index int) {
+	index = -1
 	for i := 0; i < len(nums); i++ {
 		if nums[i] == target {
-			ans = i
+			index = i
 			break
 		}
 	}
--- a/codes/go/chapter_array_and_linkedlist/array.go
+++ b/codes/go/chapter_array_and_linkedlist/array.go
@@ -1,3 +1,7 @@
+// File: array.go
+// Created Time: 2022-12-29
+// Author: GuoWei (gongguowei01@gmail.com)
+
 package chapter_array_and_linkedlist
 
 import (
--- a/codes/go/chapter_array_and_linkedlist/array.go
+++ b/codes/go/chapter_array_and_linkedlist/array.go
@@ -1,0 +1,83 @@
+package chapter_array_and_linkedlist
+
+import (
+	"fmt"
+	"math/rand"
+)
+
+// ExpectSize 预期大小。 在 Go 中，声明数组长度必须是常量表达式，
+// 所以这里我们约定，扩展后的长度为 6
+const ExpectSize = 6
+
+/* 随机返回一个数组元素 */
+func randomAccess(nums [5]int) (ans int) {
+	// 在区间 [0, nums.length) 中随机抽取一个数字
+	randomIndex := rand.Intn(len(nums))
+	// 获取并返回随机元素
+	ans = nums[randomIndex]
+	return
+}
+
+/* 扩展数组长度 */
+func extend(nums [5]int) [ExpectSize]int {
+	// 初始化一个扩展长度后的数组
+	var res [ExpectSize]int
+	// 将原数组中的所有元素复制到新数组
+	for i := 0; i < len(nums); i++ {
+		res[i] = nums[i]
+	}
+	// 返回扩展后的新数组
+	return res
+}
+
+/* 在数组的索引 index 处插入元素 num */
+func insert(nums *[5]int, num int, index int) {
+	// 把索引 index 以及之后的所有元素向后移动一位
+	// 如果超出了数组长度，会被直接舍弃
+	for i := len(nums) - 1; i > index; i-- {
+		nums[i] = nums[i-1]
+	}
+	// 将 num 赋给 index 处元素
+	nums[index] = num
+}
+
+/* 删除索引 index 处元素 */
+func remove(nums *[5]int, index int) {
+	// 越界检查
+	if index >= len(nums) {
+		return
+	}
+	// 把索引 index 之后的所有元素向前移动一位
+	for i := index; i < len(nums); i++ {
+		if i+1 >= len(nums) {
+			nums[len(nums)-1] = 0
+			break
+		}
+		nums[i] = nums[i+1]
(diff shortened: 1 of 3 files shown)


guowei-gong:
@krahets

codes/go/chapter_array_and_linkedlist/array_test.go
```diff
@@ -0,0 +1,145 @@
...
+	ans := randomAccess(nums)
+	if ans < min || ans > max {
+		t.Fatalf("Expected range is greater than min: %v and less than max: %v, got ans: %v", min, max, ans)
+	}
+}
+
+func TestExtend(t *testing.T) {
+	t.Parallel()
+	nums := [5]int{1, 2, 3, 4, 5}
+	newNums := extend(nums)
```
krahets, line 31: Can we add an argument of `enlarge` to this function `extend(nums, enlarge)` to keep the consistency with other languages?

krahets, line 31: Can we add an argument of `enlarge` to this function `extend(nums, enlarge)` to keep the consistency with other languages?

codes/go/chapter_array_and_linkedlist/array.go
```diff
@@ -0,0 +1,83 @@
...
+/* 删除索引 index 处元素 */
+func remove(nums *[5]int, index int) {
+	// 越界检查
+	if index >= len(nums) {
+		return
+	}
+	// 把索引 index 之后的所有元素向前移动一位
+	for i := index; i < len(nums); i++ {
+		if i+1 >= len(nums) {
+			nums[len(nums)-1] = 0
```
krahets, line 53: What is this used for?

codes/go/chapter_array_and_linkedlist/array.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	// 将 num 赋给 index 处元素
+	nums[index] = num
+}
+
+/* 删除索引 index 处元素 */
+func remove(nums *[5]int, index int) {
+	// 越界检查
+	if index >= len(nums) {
+		return
```
krahets, line 48: Not necessary because the for loop will handle this.

codes/go/chapter_array_and_linkedlist/array.go
```diff
@@ -0,0 +1,83 @@
...
+func randomAccess(nums [5]int) (ans int) {
+	// 在区间 [0, nums.length) 中随机抽取一个数字
+	randomIndex := rand.Intn(len(nums))
+	// 获取并返回随机元素
+	ans = nums[randomIndex]
+	return
+}
+
+/* 扩展数组长度 */
+func extend(nums [5]int) [ExpectSize]int {
```
krahets, line 22: extend(nums, enlarge) has two arguments.

codes/go/chapter_array_and_linkedlist/array.go
```diff
@@ -0,0 +1,83 @@
...
+// 所以这里我们约定，扩展后的长度为 6
+const ExpectSize = 6
+
+/* 随机返回一个数组元素 */
+func randomAccess(nums [5]int) (ans int) {
+	// 在区间 [0, nums.length) 中随机抽取一个数字
+	randomIndex := rand.Intn(len(nums))
+	// 获取并返回随机元素
+	ans = nums[randomIndex]
+	return
```
krahets, line 18: Is this line should be `return randomNum`?

codes/go/chapter_array_and_linkedlist/array.go
```diff
@@ -0,0 +1,83 @@
...
+// ExpectSize 预期大小。 在 Go 中，声明数组长度必须是常量表达式，
+// 所以这里我们约定，扩展后的长度为 6
+const ExpectSize = 6
+
+/* 随机返回一个数组元素 */
+func randomAccess(nums [5]int) (ans int) {
+	// 在区间 [0, nums.length) 中随机抽取一个数字
+	randomIndex := rand.Intn(len(nums))
+	// 获取并返回随机元素
+	ans = nums[randomIndex]
```
krahets, line 17: Replace `ans` with `randomNum`. Please keep the naming consistency with the reference language.

codes/go/chapter_array_and_linkedlist/array.go
```diff
@@ -0,0 +1,83 @@
+package chapter_array_and_linkedlist
+
+import (
+	"fmt"
+	"math/rand"
+)
+
+// ExpectSize 预期大小。 在 Go 中，声明数组长度必须是常量表达式，
+// 所以这里我们约定，扩展后的长度为 6
+const ExpectSize = 6
```
krahets, line 10: We use an enlarge argument.

krahets (commented):
Thanks for the professional code! 👍 

However, this book aims to introduce data structure and algorithms. Thus it is recommended to make the test code simpler, ignoring part of professionalism. Could you refer to other Go codes and make the test code like them?

Moreover, please check the correspondence between the code with the reference code and make them as similar as possible. It will decrease some annoying problems caused by the mismatches.

codes/go/chapter_array_and_linkedlist/array.go
guowei-gong: "The length is part of the array's type and must be a constant
expression that evaluates to a non-negative integer value." enlarge argument is
variable, not a constant expression.
https://go.dev/ref/spec#Array_types

guowei-gong:
"The length is part of the array's type and must be a constant
expression that evaluates to a non-negative integer value." n is
variable, not a constant expression. 

https://go.dev/ref/spec#Array_types

So I have to use constants instead of passing variable parameters in the function signature.  If I use the passed parameters, this will be a slice, not an array. 

@krahets 😄

codes/go/chapter_array_and_linkedlist/array.go
guowei-gong: ok, thank u 😄

codes/go/chapter_array_and_linkedlist/array.go
guowei-gong: Please see my comments below https://github.com/krahets/hello-algo/pull/178#issuecomment-1365704881

codes/go/chapter_array_and_linkedlist/array.go
guowei-gong: Please see my comments below https://github.com/krahets/hello-algo/pull/178#issuecomment-1365704881

codes/go/chapter_array_and_linkedlist/array_test.go
guowei-gong: Please see my comments below https://github.com/krahets/hello-algo/pull/178#issuecomment-1365704881

guowei-gong: Please see my comments below https://github.com/krahets/hello-algo/pull/178#issuecomment-1365704881

codes/go/chapter_array_and_linkedlist/array.go
guowei-gong: If the passed parameter index is greater than the array length - 1, the last element of the array is always modified. If not, it may not be as expected.

krahets: Why does the "the last element of the array is always modified"?

For example, if `i == index`, will the loop `for i := index; i < len(nums); i++ ` be executed?

guowei-gong:
> Thanks for the professional code! 👍
> 
> However, this book aims to introduce data structure and algorithms. Thus it is recommended to make the test code simpler, ignoring part of professionalism. Could you refer to other Go codes and make the test code like them?
> 
> Moreover, please check the correspondence between the code with the reference code and make them as similar as possible. It will decrease some annoying problems caused by the mismatches.

Thank you for your review. 😄 
I will try to modify it. I checked other people's code. The test case is a hybrid pattern. Are you sure this is the desired result? If you use the go programming style, the current pr is better. Maybe I can create another test case with mixed patterns, while keeping the files in the current pr. It's up to you, friend.

krahets:
The test case is a hybrid pattern. Are you sure this is the desired result?

Yes. Let's make the code simpler. We want the readers to focus on the algorithm.

krahets:
So I have to use constants instead of passing variable parameters in the function signature. If I use the passed parameters, this will be a slice, not an array.

If so, let's use a slice. And add a comment "我们将 Go 中的 slice 看作是数组"。

guowei-gong:
@krahets hi, I finished

codes/go/chapter_array_and_linkedlist/array.go
krahets: @guowei-gong Hi, please address this comment

codes/go/chapter_array_and_linkedlist/array_test.go
```diff
@@ -0,0 +1,43 @@
+package chapter_array_and_linkedlist
```
krahets, line 1: Add a file header. You can refer to other go files.

codes/go/chapter_array_and_linkedlist/array.go
```diff
@@ -0,0 +1,80 @@
+package chapter_array_and_linkedlist
```
krahets, line 1: Add a file header. You can refer to other go files.

krahets (commented):
Hi @guowei-gong , there are several comments to be addressed.

codes/go/chapter_array_and_linkedlist/array.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	// 将 num 赋给 index 处元素
+	nums[index] = num
+}
+
+/* 删除索引 index 处元素 */
+func remove(nums []int, index int) {
+	// 把索引 index 之后的所有元素向前移动一位
+	for i := index; i < len(nums); i++ {
+		if i+1 >= len(nums) {
```
krahets, line 49: Is this `if i+1 >= len(nums)` necessary?

krahets (commented):
There still a comment to be addressed

https://github.com/krahets/hello-algo/pull/178/files#r1058474851

guowei-gong:
> There still a comment to be addressed
> 
> https://github.com/krahets/hello-algo/pull/178/files#r1058474851

Yes, it is a must. If I now have an array of length 5, then the max index should be 4, without it, index 5 will be accessed, which will generate an error.panic: runtime error: index out of range [5] with length 5

krahets (approved):
Thank @guowei-gong for the hard work!

krahets (approved):
Fine tuned.

krahets:
> > There still a comment to be addressed
> > https://github.com/krahets/hello-algo/pull/178/files#r1058474851
> 
> Yes, it is a must. If I now have an array of length 5, then the max index should be 4, without it, index 5 will be accessed, which will generate an error.panic: runtime error: index out of range [5] with length 5

It is recommended to directly write as

```go
/* 删除索引 index 处元素 */
func remove(nums []int, index int) {
	// 把索引 index 之后的所有元素向前移动一位
	for i := index; i < len(nums) - 1; i++ {
		nums[i] = nums[i+1]
	}
}
```

Because we did not set 0 as the `empty`, we do not have to change the tail number as 0.
Moreover, if you change the code from the reference language (like here setting the tail number as 0), please describe the reason in the PR. We aim to keep the consistency of the codes of all the languages.

PR #183, opened 2022-12-28
docs(binary_tree): fix comment style and punctuation

guowei-gong:
According to the previous annotation style, we should use `/* ... */`.

`二叉树中度的范围是 0, 1, 2 `，修改为 `二叉树中，度的范围是 0, 1, 2`，如果不添加 `，`，会误导读者有一个`中度`的概念。

btw, should the style of punctuation be unified? Full width symbol or half width symbol.


krahets (approved):
Thanks for the polishing!

PR #185, opened 2022-12-28
docs(binary_search_tree): variable name error

guowei-gong:
Combining Context, Compare 'cur.val' and 'num', not 'cur.val' and 'val`.


krahets (approved):
Thanks for the correction!

PR #216, opened 2023-01-07
Add the TypeScript code and docs for Chapter of Linear Search

danielsss:
added: linear_search.ts and arrToLinkedList function
updated: linear search docs for Typescript


.editorconfig
```diff
@@ -0,0 +1,14 @@
+# EditorConfig is awesome: https://EditorConfig.org
```
krahets, line 1: Is this an editor-related file or a language-related file?

krahets (commented):
Thanks! Waiting the final review by @justin-tse

.editorconfig
danielsss: It's just an editor-related file.

krahets: I see. Please remove it. We should not keep any editor-related files in the repo.

danielsss: done

justin-tse (approved):
Thank @danielsss for the clean codes!

krahets (approved):

PR #223, opened 2023-01-08
feat: update Dockerfile to deploy

xiongsp:

This PR is not related to coding or code translation. This PR solves #134.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.9.0-buster
+FROM python:3.9.0-alpine
 
 RUN pip install -i https://pypi.tuna.tsinghua.edu.cn/simple --upgrade pip && pip install -i https://pypi.tuna.tsinghua.edu.cn/simple mkdocs mkdocs-material
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,12 @@
 FROM python:3.9.0-alpine
 
-RUN pip install -i https://pypi.tuna.tsinghua.edu.cn/simple --upgrade pip && pip install -i https://pypi.tuna.tsinghua.edu.cn/simple mkdocs mkdocs-material
+RUN pip install -i https://pypi.tuna.tsinghua.edu.cn/simple --upgrade pip
+RUN pip install -i https://pypi.tuna.tsinghua.edu.cn/simple mkdocs-material==9.0.2
 
 WORKDIR /app
 
 COPY codes /app/codes
-
 COPY docs /app/docs
-
 COPY mkdocs.yml /app/mkdocs.yml
 
 RUN mkdir ./docs/overrides && mkdocs build
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM python:3.9.0-buster
+
+RUN sed -i 's/deb.debian.org/mirrors.aliyun.com/g' /etc/apt/sources.list
+RUN sed -i 's/security.debian.org/mirrors.aliyun.com/g' /etc/apt/sources.list
+
+RUN pip install --upgrade pip && pip install mkdocs mkdocs-material
+
+WORKDIR /app
+
+COPY codes /app/codes
+
+COPY docs /app/docs
+
+COPY mkdocs.yml /app/mkdocs.yml
+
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM python:3.9.0-buster
+
+RUN sed -i 's/deb.debian.org/mirrors.aliyun.com/g' /etc/apt/sources.list
+RUN sed -i 's/security.debian.org/mirrors.aliyun.com/g' /etc/apt/sources.list
+
+RUN pip install --upgrade pip && pip install mkdocs mkdocs-material
+
+WORKDIR /app
+
+COPY codes /app/codes
```
krahets, line 10: Maybe `Add . /app` is clearer.

Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM python:3.9.0-buster
+
+RUN sed -i 's/deb.debian.org/mirrors.aliyun.com/g' /etc/apt/sources.list
```
krahets, line 3: Is this setting only for Linux? How about Win and macOS?

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN sed -i 's/security.debian.org/mirrors.aliyun.com/g' /etc/apt/sources.list
+
+RUN pip install --upgrade pip && pip install mkdocs mkdocs-material
+
+WORKDIR /app
```
krahets, line 8: Could we use `hello-algo` instead of `app`?

krahets (commented):
Thank you for adding the dockerfile. Please address the comments.

Dockerfile
xiongsp: If use `COPY . /app`, other files like `LICENSE` or `.gitignore` will be copied into the **DOCKER**, which is unnecessary.

Dockerfile
xiongsp: NO. `RUN` means **to run in the docker** but not the **host**. So there is no compatible problem. But here I just made a mistake, I will correct it later.

Dockerfile
xiongsp: YES, we can. But it is unnecessary for it is stored inside the **docker**. Nobody can see its name unless use `docker exec -it xxx` to get a shell in the docker. `app` is the tradition name for python service.

xiongsp:
If necessary, how about adding instruction in `README` or somewhere else guiding users to deploy the docker?

Dockerfile
xiongsp: Now it works correctly.

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM python:3.9.0-alpine
```
xiongsp, line 1: This aims to reduce the size of images from ~1G to ~180M

README.md
```diff
@@ -41,6 +41,28 @@
 
 如果感觉本书对你有所帮助，请点个 Star :star: 支持一下，谢谢！
 
+### 本地部署
```
krahets, line 44: Please move the tutorial to `docs/chapter_preface/contribution.md`. Docker is not for most readers.

krahets (commented):
Thanks, @xiongsp! Please address the comment. I've modified the Dockerfile. Could you help test it?

xiongsp:
@krahets, I noticed that you split up the RUN command. Although it would increase the size of images, it works and the increasement of the size is bearable. The Dockerfile has been tested and it works correctly. And the tutorial has been moved to `docs`.

Dockerfile
krahets: Great change.

krahets (approved):
Thanks for the great PR!

krahets:
> @krahets, I noticed that you split up the RUN command. Although it would increase the size of images, it works and the increasement of the size is bearable. The Dockerfile has been tested and it works correctly. And the tutorial has been moved to `docs`.

I see. I'll merge them.

krahets:
> @krahets, I noticed that you split up the RUN command. Although it would increase the size of images, it works and the increasement of the size is bearable. The Dockerfile has been tested and it works correctly. And the tutorial has been moved to `docs`.

I see. I'll merge them.

PR #237, opened 2023-01-11
add zig codes for Section 'Array', 'LinkedList' and  'List'

sjinzh:

### DIFF
--- a/codes/zig/chapter_array_and_linkedlist/array.zig
+++ b/codes/zig/chapter_array_and_linkedlist/array.zig
@@ -90,10 +90,6 @@ pub fn find(nums: []i32, target: i32) i32 {
 
 // Driver Code
 pub fn main() !void {
-    // 查看本地CPU架构和操作系统信息
-    var native_target_info = try std.zig.system.NativeTargetInfo.detect(std.zig.CrossTarget{});
-    std.debug.print("Native Info: CPU Arch = {}, OS = {}\n", .{native_target_info.target.cpu.arch, native_target_info.target.os.tag});
-
     // 初始化数组
     const size: i32 = 5;
     var arr = [_]i32{0} ** size;
--- a/codes/zig/chapter_array_and_linkedlist/array.zig
+++ b/codes/zig/chapter_array_and_linkedlist/array.zig
@@ -137,7 +137,7 @@ pub fn main() !void {
 
     // 查找元素
     var index = find(nums, 3);
-    std.debug.print("\n在 nums_ext 中查找元素 3 ，得到索引 = {}\n", .{index});
+    std.debug.print("\n在 nums 中查找元素 3 ，得到索引 = {}\n", .{index});
 
     const getchar = try std.io.getStdIn().reader().readByte();
     _ = getchar;
--- a/codes/zig/chapter_array_and_linkedlist/array.zig
+++ b/codes/zig/chapter_array_and_linkedlist/array.zig
@@ -1,0 +1,149 @@
+// File: array.zig
+// Created Time: 2023-01-07
+// Author: sjinzh (sjinzh@gmail.com)
+
+const std = @import("std");
+const inc = @import("include");
+
+// 随机返回一个数组元素
+pub fn randomAccess(nums: []i32) i32 {
+    // 在区间 [0, nums.len) 中随机抽取一个整数
+    var randomIndex = std.crypto.random.intRangeLessThan(usize, 0, nums.len);
+    // 获取并返回随机元素
+    var randomNum = nums[randomIndex];
+    return randomNum;
+}
+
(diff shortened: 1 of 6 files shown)


sjinzh:
The output of "chapter_array_and_linkedlist/array.zig" as shown below,


The output of "chapter_array_and_linkedlist/linked_list.zig" as shown below,


The output of "chapter_array_and_linkedlist/list.zig" as shown below,


The output of "chapter_array_and_linkedlist/my_list.zig" as shown below,

codes/zig/chapter_array_and_linkedlist/array.zig
```diff
@@ -0,0 +1,145 @@
...
+    // 将原数组中的所有元素复制到新数组
+    std.mem.copy(i32, res, nums);
+    // 返回扩展后的新数组
+    return res;
+}
+
+// 扩展数组长度（编译期方法A：初始化新数组并进行元素拷贝）
+pub fn extendComptimeA(comptime nums: anytype, comptime enlarge: i32) [nums.len + enlarge]i32 {
+    // 初始化一个扩展长度后的数组
+    var res = [_]i32{0} ** (nums.len + enlarge);
```
krahets, line 31: Is the length of this array determined in the compilation? If so, why could we use the non-constant number to set its length?

codes/zig/chapter_array_and_linkedlist/my_list.zig
```diff
@@ -0,0 +1,178 @@
...
+        pub fn toArray(self: *Self) ![]T {
+            // 仅转换有效长度范围内的列表元素
+            var nums = try self.mem_allocator.alloc(T, self.size());
+            std.mem.set(T, nums, @as(T, 0));
+            for (nums) |*num, i| {
+                num.* = self.get(i);
+            }
+            return nums;
+        }
+
```
krahets, line 123: Remove the empty line.

krahets (commented):
Thanks for the great work! Please address the comment.

codes/zig/chapter_array_and_linkedlist/my_list.zig
sjinzh: > Remove the empty line.

done.

codes/zig/chapter_array_and_linkedlist/array.zig
sjinzh: > Is the length of this array determined in the compilation? If so, why could we use the non-constant number to set its length?

It's a good question! 
"In Zig, the programmer can label variables as comptime. This guarantees to the compiler that every load and store of the variable is performed at compile-time. ". For more details, please see Zig Language Reference(https://ziglang.org/documentation/0.10.0/).

krahets: Good explanation! More 2 questions:

1. Is `var res = [_]i32{0} ** (nums.len + enlarge);` this array be stored on stack or heap?
2. It is worth discussing the logical correctness to extend array size during compile-time. "comptime variables" means that "we already know the array should be extended", thus why don't we directly create an array with sufficient size?

sjinzh: > Good explanation! More 2 questions:
> 
> 1. Is `var res = [_]i32{0} ** (nums.len + enlarge);` this array be stored on stack or heap?
> 2. It is worth discussing the logical correctness to extend array size during compile-time. "comptime variables" means that "we already know the array should be extended", thus why don't we directly create an array with sufficient size?

1. Refer to https://ziglang.org/documentation/0.10.0/#Where-are-the-bytes, "Just like string literals, const declarations, when the value is known at [comptime](https://ziglang.org/documentation/0.10.0/#comptime), are stored in the global constant data section. Also [Compile Time Variables](https://ziglang.org/documentation/0.10.0/#Compile-Time-Variables) are stored in the global constant data section.". So i think the right part "[_]i32{0} ** (nums.len + enlarge)" is in the global constant data section,  because "comptime nums" and "comptime enlarge" , and the left part "res" is in stack data section.
2. Yes, we can directly create an array with sufficient size if we know the array should be extended. Furthermore, if we already know the size of expanded array, we can specify the exact size in comptime for memory saving especially running in the embedded devices.

krahets: I see. I recommend deleting it. As far as we discussed, the "extending size for array" is a concept only meaningful for the runtime. In other words, we will never extend the array size in compile-time.

sjinzh: > I see. I recommend deleting it. As far as we discussed, the "extending size for array" is a concept only meaningful for the runtime. In other words, we will never extend the array size in compile-time.

done.

krahets (approved):

PR #196, opened 2023-01-01
code: added Typescript and Javascript examples

RiverTwilight:
Not sure whether these formats meet the requirement. If everything is okay I will continue to transcribe more:-)

If this PR is related to coding or code translation, please fill out the checklist.

- [x] I've tested the code and ensured the outputs are the same as the outputs of reference codes.
- [x] I've checked the codes (formatting, comments, indentation, file header, etc) carefully.
- [x] The code does not rely on a particular environment or IDE and can be executed on a standard system (Win, macOS, Ubuntu).

PR #245, opened 2023-01-12
ignore markdown for prettier

RiverTwilight:
During my previous [PR](https://github.com/krahets/hello-algo/pull/196#issuecomment-1368966563), I found that prettier will automatically add three spaces after `-` for each list item, which doesn't match the project's format.

In Prettier's rules, only the `--prose-wrap` option is available for markdown([credit](https://prettier.io/blog/2017/11/07/1.8.0.html)). so I think maybe we can just ignore the markdown file, or adapt prettier's rules to all files. 

This change is mainly for the sake of those who enabled the "**Format On Save**" option in VSCode and those who like to immediately format after each change like me.


krahets:
Thanks for the clear description! I think we should not upload IDE-related files. It is a case-by-case setting for the users.

PR #259, opened 2023-01-15
fix the Golang code for median in quick_sort.md

dshlstarr:
Fix the Golang code for median

### DIFF
--- a/docs/chapter_sorting/quick_sort.md
+++ b/docs/chapter_sorting/quick_sort.md
@@ -483,7 +483,7 @@ comments: true
     func medianThree(nums []int, left, mid, right int) int {
         if (nums[left] > nums[mid]) != (nums[left] > nums[right]) {
             return left
-        } else if (nums[mid] < nums[left]) != (nums[mid] > nums[right]) {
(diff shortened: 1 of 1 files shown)


docs/chapter_sorting/quick_sort.md
```diff
@@ -483,7 +483,7 @@ comments: true
     func medianThree(nums []int, left, mid, right int) int {
         if (nums[left] > nums[mid]) != (nums[left] > nums[right]) {
             return left
-        } else if (nums[mid] < nums[left]) != (nums[mid] > nums[right]) {
+        } else if (nums[mid] > nums[left]) != (nums[mid] > nums[right]) {
```
Reanon, line 486: There is no error, you can test the code.

krahets (approved):
Thanks for the fix!

docs/chapter_sorting/quick_sort.md
Reanon, line 486: It seems that this problem is also found in other code, you can modify the codes meanwhile.

krahets: > It seems that this problem is also found in other code, you can modify the codes meanwhile.

No need to fix other codes.

PR #261, opened 2023-01-15
feat(heap): add c codes.

Reanon:


codes/c/chapter_heap/my_heap.c
```diff
@@ -0,0 +1,179 @@
...
+
+// 大顶堆
+typedef struct maxHeap {
+    // size 代表的是实际元素的个数
+    int size;
+    // 使用预先分配内存的数组，避免扩容
+    int data[MAX_SIZE];
+} maxHeap;
+
+void siftDown(maxHeap *h, int i);
```
krahets, line 19: Do we need it?

codes/c/chapter_heap/my_heap.c
```diff
@@ -0,0 +1,179 @@
...
+
+void siftUp(maxHeap *h, int i);
```
krahets, line 21: Do we need it?

codes/c/chapter_heap/my_heap.c
```diff
@@ -0,0 +1,179 @@
...
+    printHeap(heap->data, heap->size);
+
+    /* 获取堆大小 */
+    printf("\n堆元素数量为 %d\n", size(heap));
+
+    /* 判断堆是否为空 */
+    printf("\n堆是否为空 %d\n", isEmpty(heap));
+
+    // 释放内存
+    free(heap);
```
krahets, line 178: Use delete instead.

> **Differences between delete and free()**
> The delete operator is used to delete the pointer, which is either allocated using new operator or a NULL pointer, whereas the free() function is used to delete the pointer that is either allocated using malloc(), calloc() or realloc() function or NULL pointer.

codes/c/include/print_util.h
```diff
@@ -25,7 +25,7 @@ extern "C" {
  * @param arr
  * @param size
  */
-static void printArray(int *arr, int size) {
+static void printArray(int arr[], int size) {
```
krahets, line 28: This modification influence other files such as `array.c`. Have you tested them?

codes/c/chapter_heap/my_heap.c
```diff
@@ -0,0 +1,179 @@
...
+        // 交换两结点
+        swap(h, i, max);
+        // 循环向下堆化
+        i = max;
+    }
+}
+
+/* 从结点 i 开始，从底至顶堆化 */
+void siftUp(maxHeap *h, int i) {
+    int p;
```
krahets, line 135: You can initialize the `p` in the while loop. It will make no difference in memory footprint but make the code simpler.

https://stackoverflow.com/questions/8735765/variables-declared-inside-a-loop

codes/c/chapter_heap/my_heap.c
```diff
@@ -0,0 +1,179 @@
...
+    siftDown(h, 0);
+
+    // 返回堆顶元素
+    return val;
+}
+
+
+/* 从结点 i 开始，从顶至底堆化 */
+void siftDown(maxHeap *h, int i) {
+    int l, r, max;
```
krahets, line 110: Same as below.

codes/c/chapter_heap/my_heap.c
```diff
@@ -0,0 +1,179 @@
...
+    // 使用预先分配内存的数组，避免扩容
+    int data[MAX_SIZE];
+} maxHeap;
+
+void siftDown(maxHeap *h, int i);
+
+void siftUp(maxHeap *h, int i);
+
+/* 构造函数，根据切片建堆 */
+maxHeap *newMaxHeap(int nums[], int size) {
```
krahets, line 24: We also need a constructor to initialize an empty heap.

krahets (commented):
Thanks! Please address the comment.

codes/c/chapter_heap/my_heap.c
Reanon: There is no delete in C.

codes/c/include/print_util.h
Reanon: I have tested, it works as well.

codes/c/chapter_heap/my_heap.c
Reanon: Modified in next commit.

codes/c/chapter_heap/my_heap.c
Reanon: Yes.A function can't be called before defining it in C. If not  declared here, we should move "newMaxHeap" below "siftDown".

krahets (approved):
Thanks!

PR #272, opened 2023-01-18
fix: 更快 -> 较快

FangYuan33:
较大、较慢、较快、较小... 这样顺一些


krahets (approved):
Thanks for the polishing!

PR #286, opened 2023-01-19
fix: 二分查找，左闭右开区间实现中，注释区间范围右侧符号应该是)而不应该是]

FangYuan33:
我觉得需要改一下，这样更好理解


krahets (commented):
Thanks for the fix! Could you help update the comments in the source code files as well?

FangYuan33:
> Thanks for the fix! Could you help update the comments in the source code files as well?

ok! my pleasure. when fixing it, I lost code need...

FangYuan33:
> Thanks for the fix! Could you help update the comments in the source code files as well?

Only TypeScript is wrong, others is right.

krahets (approved):
Great! Now merge it.

PR #288, opened 2023-01-19
Fix unknown behavior when size is 0 or arr is a null pointer in C header file

Gonglja:
…inter

### DIFF
--- a/codes/c/include/print_util.h
+++ b/codes/c/include/print_util.h
@@ -27,17 +27,21 @@ extern "C" {
  */
 static void printArray(int *arr, int size) {
     printf("[");
-    for (int i = 0; i < size - 1; i++) {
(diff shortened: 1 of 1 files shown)


codes/c/include/print_util.h
```diff
@@ -27,17 +27,21 @@ extern "C" {
...
+      for (int i = 0; i < size - 1; i++) {
+          if (arr[i] != NIL) {
+              printf("%d, ", arr[i]);
+          } else {
+              printf("NULL, ");
+          }
+      }
+      if (arr[size - 1] != NIL) {
+          printf("%d]\n", arr[size - 1]);
+      }else{
```
krahets, line 40: } else {

krahets (approved):
Thanks for the fix!

PR #294, opened 2023-01-21
add Rust code array, list

xBLACKICEx:

array output:
image

list: output
image


krahets (approved):
Thanks for the PR! Happy Chinese new year!

PR #321, opened 2023-02-04
Part Heap : Code Translation into C++ (my_heap.cpp)

longranger2:
If this PR is related to coding or code translation, please fill out the checklist.

 ✅  I've tested the code and ensured the outputs are the same as the outputs of reference codes.
 ✅  I've checked the codes (formatting, comments, indentation, file header, etc) carefully.
 ✅  The code does not rely on a particular environment or IDE and can be executed on a standard system (Win, macOS, Ubuntu).

我已经完成了my_heap.cpp，实现效果如下：
```C++
输入数组并建堆后
堆的数组表示：
[9, 8, 6, 6, 7, 5, 2, 1, 4, 3, 6, 2]
堆的树状表示：
        /——— 2
    /——— 6
   |    \——— 5
   |        \——— 2
——— 9
   |        /——— 6
   |    /——— 7
   |   |    \——— 3
    \——— 8
       |    /——— 4
        \——— 6
            \——— 1
堆顶元素为 9
元素 7 入堆后
堆的数组表示：
[9, 8, 7, 6, 7, 6, 2, 1, 4, 3, 6, 2, 5]
堆的树状表示：
        /——— 2
    /——— 7
   |   |    /——— 5
   |    \——— 6
   |        \——— 2
——— 9
   |        /——— 6
   |    /——— 7
   |   |    \——— 3
    \——— 8
       |    /——— 4
        \——— 6
            \——— 1
堆顶元素 7 出堆后
堆的数组表示：
[8, 7, 7, 6, 6, 6, 2, 1, 4, 3, 5, 2]
堆的树状表示：
        /——— 2
    /——— 7
   |    \——— 6
   |        \——— 2
——— 8
   |        /——— 5
   |    /——— 6
   |   |    \——— 3
    \——— 7
       |    /——— 4
        \——— 6
            \——— 1
堆顶元素数量为 12
堆是否为空 0
```
image

PR #346, opened 2023-02-06
fix the pointer recycle problem of heap.cpp and my_heap.cpp

longranger2:
Fix the problems mentioned in the PR https://github.com/krahets/hello-algo/pull/321


krahets (approved):
Thanks for the quick fix!

Issue #15, opened 2022-11-25
Codes Translation Help wanted 


krahets:
We're wondering about translating codes from language A to language B.

The data structures and algorithms are first implemented and fully tested in Java. Therefore, it is recommended to regard Java codes as a coding style standard.

- [x] **Java** by @krahets 
- [ ] **C**
- [x] **C++** by @krahets (Can be replaced)
- [ ] **C#**
- [x] **Python** by @krahets (Can be replaced)
- [x] **Golang** by @Reanon 
- [ ] **JavaScript**
- [ ] **TypeScript**

Welcome translation to more unlisted languages.

kalaersun:
I can help translate part of C

krahets:
> I can help translate part of C

Thanks, @kalaersun ! Looking forward for your code.

lkgGitHub:
I can help translate part of go about "Array and Linked List" and "Sorting"

krahets:
> I can help translate part of go about "Array and Linked List" and "Sorting"

Thanks, @lkgGitHub !
Please get in touch with @Reanon cause maybe he is working on the remaining Go codes.

Reanon:
> I can help translate part of go about "Array and Linked List" and "Sorting"

@lkgGitHub, You can do it! Chapter Array/LinkedList or Sorting are available. If you pull request, i will review you golang code. Beside, before you pull request, you should check if other pull request to Golang code. Thank you for participating in Golang Translation😁.

justin-tse:
I can help translate part of TypeScript

justin-tse:
> 

Computational Complexity， Array and Linked List，Stack and Queue

justin-tse:
I will try to translate the part of Array and Linked List.

justin-tse:
I think I can translate the remaining part of Chapter Array and LinkedList to JavaScript.

msk397:
I can help translate the part of Insertion Sort and Quick Sort to Go

krahets:
> I think I can translate the remaining part of Chapter Array and LinkedList to JavaScript.

Hi @justin-tse , the JS code of the Chapter Array and LinkedList has been committed in #93

mingXta:
I can help translate part of C# start with Array.

Cathay-Chen:
I can help translate part of Go

justin-tse:
I can translate the Chapter of Tree to TypeScript.

msk397:
[...]

PR #357, opened 2023-02-09
Add graph c++ code

what-is-me:

### DIFF
--- a/codes/cpp/chapter_graph/graph_adjacency_list.cpp
+++ b/codes/cpp/chapter_graph/graph_adjacency_list.cpp
@@ -1,0 +1,126 @@
+/**
+ * File: graph_adjacency_matrix.java
+ * Created Time: 2023-01-26
+ * Author: what-is-me (whatisme@outlook.jp)
+ */
+
+#include <unordered_set>
+
+#include "../include/include.hpp"
+
+/* 顶点类 */
+struct Vertex {
+    int val;
+    Vertex(int val) : val(val) {}
+    bool operator==(const Vertex &v) const { return v.val == val; }
+};
+
+/* 顶点类的哈希类 */
+namespace std {
+    template <>
+    class hash<Vertex> {
+    public:
+        size_t operator()(const Vertex &vertex) const { return vertex.val; }
+    };
+}
+/* 基于邻接表实现的无向图类 */
+class GraphAdjList {
+    // 请注意，vertices 和 adjList 中存储的都是 Vertex 对象
+    unordered_map<Vertex, unordered_set<Vertex>>
+        adjList;  // 邻接表（使用哈希表实现）
+
+public:
+    /* 构造函数 */
+    GraphAdjList(const vector<vector<Vertex>> &edges) {
+        // 添加所有顶点和边
+        for (const vector<Vertex> &edge : edges) {
+            addVertex(edge[0]);
+            addVertex(edge[1]);
+            addEdge(edge[0], edge[1]);
+        }
+    }
+
+    /* 获取顶点数量 */
+    int size() { return adjList.size(); }
+
+    /* 添加边 */
+    void addEdge(Vertex vet1, Vertex vet2) {
+        if (!adjList.count(vet1) || !adjList.count(vet2) || vet1 == vet2)
+            throw invalid_argument("不存在顶点");
+        // 添加边 vet1 - vet2
+        adjList[vet1].insert(vet2);
+        adjList[vet2].insert(vet1);
+    }
+
+    /* 删除边 */
+    void removeEdge(Vertex vet1, Vertex vet2) {
+        if (!adjList.count(vet1) || !adjList.count(vet2) || vet1 == vet2)
+            throw invalid_argument("不存在顶点");
+        // 删除边 vet1 - vet2
+        adjList[vet1].erase(vet2);
+        adjList[vet2].erase(vet1);
+    }
+
+    /* 添加顶点 */
+    void addVertex(Vertex vet) {
+        if (adjList.count(vet)) return;
+        // 在邻接表中添加一个新链表（即 HashSet）
+        adjList[vet] = unordered_set<Vertex>();
+    }
+
+    /* 删除顶点 */
+    void removeVertex(Vertex vet) {
+        if (!adjList.count(vet))
+            throw invalid_argument("不存在顶点");
+        // 在邻接表中删除顶点 vet 对应的链表（即 HashSet）
+        adjList.erase(vet);
+        // 遍历其它顶点的链表（即 HashSet），删除所有包含 vet 的边
+        for (auto &[key, set_] : adjList) {
+            set_.erase(vet);
+        }
+    }
+
+    /* 打印邻接表 */
+    void print() {
+        cout << "邻接表 =" << endl;
+        for (auto &[key, value] : adjList) {
+            vector<int> tmp;
+            for (Vertex vertex : value)
+                tmp.push_back(vertex.val);
+            cout << key.val << ": ";
+            PrintUtil::printVector(tmp);
+        }
+    }
+};
+int main() {
+    /* 初始化无向图 */
+    Vertex v0(1), v1(3), v2(2), v3(5), v4(4);
+    vector<vector<Vertex>> edges = {{v0, v1}, {v1, v2}, {v2, v3}, {v0, v3}, {v2, v4}, {v3, v4}};
+    GraphAdjList graph(edges);
+    cout << "\n初始化后，图为" << endl;
+    graph.print();
+
+    /* 添加边 */
+    // 顶点 1, 2 即 v0, v2
+    graph.addEdge(v0, v2);
+    cout << "\n添加边 1-2 后，图为" << endl;
+    graph.print();
+
+    /* 删除边 */
+    // 顶点 1, 3 即 v0, v1
+    graph.removeEdge(v0, v1);
+    cout << "\n删除边 1-3 后，图为" << endl;
+    graph.print();
+
+    /* 添加顶点 */
+    Vertex v5(6);
+    graph.addVertex(v5);
+    cout << "\n添加顶点 6 后，图为" << endl;
+    graph.print();
+
+    /* 删除顶点 */
+    // 顶点 3 即 v1
+    graph.removeVertex(v1);
+    cout << "\n删除顶点 3 后，图为" << endl;
+    graph.print();
+}
--- a/codes/cpp/chapter_graph/graph_adjacency_list.cpp
+++ b/codes/cpp/chapter_graph/graph_adjacency_list.cpp
@@ -10,29 +10,19 @@
 struct Vertex {
     int val;
     Vertex(int val) : val(val) {}
-    bool operator==(const Vertex &v) const { return v.val == val; }
 };
-
-/* 顶点类的哈希类 */
-namespace std {
-    template <>
-    class hash<Vertex> {
-    public:
-        size_t operator()(const Vertex &vertex) const { return vertex.val; }
-    };
-}
 
 /* 基于邻接表实现的无向图类 */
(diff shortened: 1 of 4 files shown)


what-is-me:
### graph_adjacency_matrix.cpp
```

初始化后，图为
顶点列表 = [1, 3, 2, 5, 4]
邻接矩阵 =
[
  [0, 1, 1, 1, 0],        
  [1, 0, 1, 0, 0],        
  [1, 1, 0, 1, 1],        
  [1, 0, 1, 0, 1],        
  [0, 0, 1, 1, 0],        
]

添加边 1-2 后，图为       
顶点列表 = [1, 3, 2, 5, 4]
邻接矩阵 =
[
  [0, 1, 1, 1, 0],        
  [1, 0, 1, 0, 0],        
  [1, 1, 0, 1, 1],        
  [1, 0, 1, 0, 1],        
  [0, 0, 1, 1, 0],        
]

删除边 1-3 后，图为
顶点列表 = [1, 3, 2, 5, 4]
邻接矩阵 =
[
  [0, 0, 1, 1, 0],
  [0, 0, 1, 0, 0],
  [1, 1, 0, 1, 1],
  [1, 0, 1, 0, 1],
  [0, 0, 1, 1, 0],
]

添加顶点 6 后，图为
顶点列表 = [1, 3, 2, 5, 4, 6]
邻接矩阵 =
[
  [0, 0, 1, 1, 0, 0],
  [0, 0, 1, 0, 0, 0],
  [1, 1, 0, 1, 1, 0],
  [1, 0, 1, 0, 1, 0],
  [0, 0, 1, 1, 0, 0],
  [0, 0, 0, 0, 0, 0],
]

删除顶点 3 后，图为
顶点列表 = [1, 2, 5, 4, 6]
邻接矩阵 =
[
  [0, 1, 1, 0, 0],
  [1, 0, 1, 1, 0],
  [1, 1, 0, 1, 0],
  [0, 1, 1, 0, 0],
  [0, 0, 0, 0, 0],
]
```
### graph_adjacency_list.cpp
```

初始化后，图为
邻接表 =
4: [5, 2]
5: [4, 1, 2]
2: [4, 5, 3]
3: [2, 1]
1: [5, 3]

添加边 1-2 后，图为
邻接表 =
4: [5, 2]
5: [4, 1, 2]
2: [1, 4, 5, 3]
3: [2, 1]
1: [2, 5, 3]

删除边 1-3 后，图为
邻接表 =
4: [5, 2]
5: [4, 1, 2]
2: [1, 4, 5, 3]
3: [2]
1: [2, 5]

添加顶点 6 后，图为
邻接表 =
6: []
4: [5, 2]
5: [4, 1, 2]
2: [1, 4, 5, 3]
3: [2]
1: [2, 5]

删除顶点 3 后，图为
邻接表 =
6: []
4: [5, 2]
5: [4, 1, 2]
2: [1, 4, 5]
1: [2, 5]
```

what-is-me:
there are some suggestions:
- add `*.exe`, `.vscode/`, `.idea/` to .gitignore
- maybe you can repalce the class PrintUtil by overloading `operator<<` so that we can easily use `cout<<vector<<endl` to print things.

and there is a little doubt about why you create the class Vertex.
It is useless and in c++, we had to overload `operator==` and create a class for hash.
I think it can be replaced by int as in `graph_adjacency_matrix.cpp`

what-is-me:
### I didn't alter anything in `graph_operations.md`  as I noticed you are building your page in a new way.

krahets:
> add *.exe, .vscode/, .idea/ to .gitignore

They are already added in the .gitignore in the base dir.

> maybe you can repalce the class PrintUtil by overloading operator<< so that we can easily use cout<<vector<<endl to print things.

Good idea! But it may cause misunderstanding to the readers that the printer is built in the C++ official libs.

> I think it can be replaced by int as in graph_adjacency_matrix.cpp

If using int as Vertex, how can we deal with the duplicated values? When wanting to remove a node, which one should be removed?

what-is-me:
> They are already added in the .gitignore in the base dir.
Thanks!
> Good idea! But it may cause misunderstanding to the readers that the printer is built in the C++ official libs.
You are right.
> If using int as Vertex, how can we handle the duplicated values? When wanting to remove a node, which one should be removed?
So `Vertex a{.val = 1}` and `Vertex b{.val=1}` aren't a same vertex if their address aren't the same?

what-is-me:
### graph_adjacency_list.cpp
the code was altered, and the output is:
```

初始化后，图为
邻接表 =
4: [5, 2]
2: [4, 5, 3]       
3: [2, 1]
5: [1, 4, 2]       
1: [5, 3]

添加边 1-2 后，图为
邻接表 =
4: [5, 2]
2: [4, 1, 5, 3]    
3: [2, 1]
5: [1, 4, 2]       
1: [2, 5, 3]       

删除边 1-3 后，图为
邻接表 =
4: [5, 2]
2: [4, 1, 5, 3]    
3: [2]
5: [1, 4, 2]
1: [2, 5]

添加顶点 6 后，图为
邻接表 =
6: []
4: [5, 2]
2: [4, 1, 5, 3]
3: [2]
5: [1, 4, 2]
1: [2, 5]

删除顶点 3 后，图为
邻接表 =
6: []
4: [5, 2]
2: [4, 1, 5]
5: [1, 4, 2]
1: [2, 5]
```
seems the same.

Now, if Vertex a and b have the same value, but different address, they are different vertex.

docs/chapter_graph/graph_operations.md
```diff
@@ -289,7 +289,9 @@ comments: true
 === "C++"
 
     ```cpp title="graph_adjacency_list.cpp"
+    [class]{Vertex}-[func]{}
```
what-is-me, line 292: And I altered the markdown as in other markdown.

what-is-me: is that right?

krahets: Yes

codes/cpp/chapter_graph/graph_adjacency_list.cpp
```diff
@@ -0,0 +1,128 @@
...
+    /* 构造方法 */
+    GraphAdjList(const vector<vector<Vertex*>>& edges) {
+        // 添加所有顶点和边
+        for (const vector<Vertex*>& edge : edges) {
+            addVertex(edge[0]);
+            addVertex(edge[1]);
+            addEdge(edge[0], edge[1]);
+        }
+    }
+
```
krahets, line 31: Missing deconstructor ~GraphAdjList()

codes/cpp/chapter_graph/graph_adjacency_list.cpp
```diff
@@ -0,0 +1,128 @@
...
+    graph.print();
+
+    /* 删除顶点 */
+    // 顶点 3 即 v1
+    graph.removeVertex(v1);
+    cout << "\n删除顶点 3 后，图为" << endl;
+    graph.print();
+
+    /* 释放内存 */
+    delete v0;
```
krahets, line 122: Not needed because deconstructor will delete them.

codes/cpp/chapter_graph/graph_adjacency_matrix.cpp
```diff
@@ -0,0 +1,124 @@
...
+        for (int val : vertices) {
+            addVertex(val);
+        }
+        // 添加边
+        // 请注意，edges 元素代表顶点索引，即对应 vertices 元素索引
+        for (const vector<int>& edge : edges) {
+            addEdge(edge[0], edge[1]);
+        }
+    }
+
```
krahets, line 27: deconstructor

krahets (commented):
Hi! Please address the comments.

what-is-me:
When I remove an vertex from the graph, I should also free it?
What if you create a subgraph? 
the same to the deconstructor.

krahets:
> When I remove an vertex from the graph, I should also free it?

Yes. We assume the vertex is used by this class only.

> What if you create a subgraph?

What do you mean by creating a subgraph?

what-is-me:
> Yes. We assume the vertex is used by this class only.

I see.
> What do you mean by creating a subgraph?

When you write a new chapter about graph, there may be some concept about subgraph.
If we create a subgraph g from graph G, and g was deleted, there may be something wrong in G.

You said the vertex is used by this class only, and the Vertex are differ by their address,
so you are going to rewrite the graph class when writing the chapter relate to subgraph? like
- Bipartite graph
- Minimum Spanning Tree
 or these are not in your plan?

krahets:
> The vertices and edges of a graph that are subsets of another graph are known as a subgraph.

The node in the subgraph and graph are the same node. When you remove a node in subgraph, the node in graph is also be removed because they're the same object.

> If we create a subgraph g from graph G, and g was deleted, there may be something wrong in G.

I don't think it will cause an error if you correctly provide a remove function in subgraph. Could you show an example?

krahets:
> The vertices and edges of a graph that are subsets of another graph are known as a subgraph.

The node in the subgraph and graph are the same node. When you remove a node in subgraph, the node in graph is also be removed because they're the same object.

> If we create a subgraph g from graph G, and g was deleted, there may be something wrong in G.

I don't think it will cause an error if you correctly provide a remove function in subgraph. Could you show an example?

what-is-me:
I will add 
```cpp
    /* 析构方法 */
    ~GraphAdjList() {
        for (auto [vet, to] : adjList) delete vet;
    }
```
and add `delete vet;` to 
```cpp
    /* 删除顶点 */
    void removeVertex(Vertex* vet) {
        if (!adjList.count(vet))
            throw invalid_argument("不存在顶点");
        // 在邻接表中删除顶点 vet 对应的链表（即 HashSet）
        adjList.erase(vet);
        // 遍历其它顶点的链表（即 HashSet），删除所有包含 vet 的边
        for (auto& [key, set_] : adjList) {
            set_.erase(vet);
        }
        delete vet;
    }
```

> The node in the subgraph and graph are the same node. When you remove a node in subgraph, the node in graph is also be removed because they're the same object.

for example,
```cpp
void samplefunction(GraphAdjList& G){
    GraphAdjList g = G.createSubgraph(); // createSubgraph() may be realized in the future.
    g.print();
}
int main(){
    Vertex *v0 = new Vertex(1),
           *v1 = new Vertex(3),
           *v2 = new Vertex(2),
           *v3 = new Vertex(5),
           *v4 = new Vertex(4);
    vector<vector<Vertex*>> edges = {{v0, v1}, {v1, v2}, {v2, v3}, {v0, v3}, {v2, v4}, {v3, v4}};
    GraphAdjList G(edges);
    samplefunction(G);
    G.print();
}
```

this will go wrong.
in function samplefunction, we created a subgraph, after the function was called, g was deleted by the compiler,  vertex of g was also auto deleted. Then in G, these vertexes are missing.

Or subgraph will be an additional class?

what-is-me:
Anyhow, I added deconstructor in [fa626ad](https://github.com/krahets/hello-algo/pull/357/commits/fa626ad5f1fa729d46502b8a46600df8e3002a83)

krahets:
Great description! I think we have to remove the deconstructor to decouple Graph and Vertex.

krahets:
And maybe we should use Vertex also in the `graphAdjacencyMat` to enable the subgraph extraction.

what-is-me:
Yes, we should be really careful to memory when writing c++. 
A better way is to use smart pointers, but it is difficult for the beginers, and may also have negative impact to readability and performance.
Moreover, a bit memory leak is uninfluential, because the program will auto free all the memory when it ends. That's why most of solutions written in Algorithm Competition never free memory they applied for. For example [Trie](https://leetcode.cn/problems/implement-trie-prefix-tree/solution/shi-xian-trie-qian-zhui-shu-by-leetcode-ti500/).
But, in writing a project, that's not a good habit.

krahets (approved):
Thanks for the clear idea! Now merge it.

PR #370, opened 2023-02-15
Add vertex class for javascript and typescript

zhuoqinyue:


krahets:
Hi @zhuoqinyue ! The Vertex class is used for the adjacency list implementation. Could you update the `adjacency_list.js` and `...ts` referring to the Java code. Thanks!

zhuoqinyue:
> Hi @zhuoqinyue ! The Vertex class is used for the adjacency list implementation. Could you update the `adjacency_list.js` and `...ts` referring to the Java code. Thanks!

OK. But there will cause a problem when I write the code of `graph_bfs`.
The `module.exports` of `js` will cause other code in `adjacency_list.js` to execute. (B
ecause I need to import `adjacency_list` to finish `graph_bfs`)
Then the execution result of `adjacency_list.js` will be output first, and then `graph_bfs` will be executed.

Do you think this problem has a big impact?

krahets:
Yes. We'd better solve this. @justin-tse How do you think about it?

zhuoqinyue:
I have a opinion that we don't have to update the `adjacency_list.js`.
And just write a new adjacency_list class in the `javascript/include` folder.

krahets:
Yes. But it is not elegant because we will make a twin class.

Is there an entry point check in JS and TS? Like `if __name__ == "__main__":` in Python.

- When running this file, the test case will be executed.
- When importing this file as a module, the test case will NOT be executed.

zhuoqinyue:
It seems there no such method.

zhuoqinyue:
emm... One more solution.
Create a main function to wrap these test cases, and then don't execute the main function.
Let the user execute it by himself

krahets:
Better but not enough elegant haha!

zhuoqinyue:
I've update the `adjacency_list` @krahets

justin-tse:
> Yes. But it is not elegant because we will make a twin class.
> 
> Is there an entry point check in JS and TS? Like `if __name__ == "__main__":` in Python.
> 
> * When running this file, the test case will be executed.
> * When importing this file as a module, the test case will NOT be executed.

@zhuoqinyue Thank you for the clear codes. @krahets I think we can use [require.main](https://nodejs.org/api/modules.html#requiremain) to solve this problem?
[Node.js, require.main === module](https://stackoverflow.com/questions/45136831/node-js-require-main-module)
I have tested it, you could try it.@zhuoqinyue
JS:
image
TS is the same.

zhuoqinyue:
Cool, guys.
It works a lot in JS. But Ts does'nt have the `requrie` @justin-tse

justin-tse:
@zhuoqinyue, I am using ts-node to test my .ts file and it seems to be working well. However, I'm curious to know what kind of problems you are experiencing.

zhuoqinyue:
It works a lot in JS, but doesn't work in TS.
`ts-node` throw a TSError
image
Your `ts-node` install in the global?

zhuoqinyue:
The result tell me that `ts-node` doesn't have `require` @justin-tse

justin-tse:
Yes, I think you should install @types/node, and try it.

zhuoqinyue:
`npm i @types/node --global`
image

`ts-node` throw the same TSError again.

zhuoqinyue:
But use `npm i --save-dev @types/node`, it works successfully.

justin-tse:
> It works a lot in JS, but doesn't work in TS. `ts-node` throw a TSError image
> 
> Your `ts-node` install in the global?

Good job! I think you misunderstand my meaning that I install ts-node in global but install @types/node in the devDependencies because it is only used in our local project. Your warning message tells you what to do.

justin-tse:
> It works a lot in JS, but doesn't work in TS. `ts-node` throw a TSError image
> 
> Your `ts-node` install in the global?

@zhuoqinyue Good job! I think you misunderstand my meaning that I install ts-node in global but install @types/node in the devDependencies because it is only used in our local project. Your warning message tells you what to do.

zhuoqinyue:
Could you tell me your config about `@types/node`, `ts-node` and `typescript`, such as json, path and version.
Because I failed again by using `npm i --save-dev @types/node` . @justin-tse

justin-tse:
> But use `npm i --save-dev @types/node`, it works successfully. image
> 
> image
I see your last message which is a success. What's the problem now? The warning message?

zhuoqinyue:
> 

Fine. I found my fault.
I should `npm init -y` in folder `hello-algo` firstly, and then `npm i --save-dev @types/node`.
Maybe we should add some gitignore rules for `node_modules` `package.json`?

justin-tse:
OK, I think we don't push node_modules and package.json to this repository. These files are a little related to this project but are too big. You could give a comment before the line 
```typescript
// need to add the package @types/node contains type definitions for Node.js, npm i --save-dev @types/node
if (require.main === module)
```

zhuoqinyue:
.gitignore make git to ignore some file or some folder change, as follows: 

- before add rules `node_modules/`:
image
- after add rules `node_modules/`:
image

But git will detect the change of. gitgnore.

zhuoqinyue:
> OK, I think we don't push node_modules and package.json to this repository. These files are a little related to this project but are too big. You could give a comment before the line
> 
> ```ts
> // need to add the package @types/node contains type definitions for Node.js, npm i --save-dev @types/node
> if (require.main === module)
> ```

I've updated it. @justin-tse

codes/typescript/chapter_graph/graph_adjacency_list.ts
```diff
@@ -79,48 +76,51 @@ class GraphAdjList {
...
+// need to add the package @types/node contains type definitions for Node.js, npm i --save-dev @types/node
+if (require.main === module) {
+    /* Driver Code */
+    /* 初始化无向图 */
+    const v0 = new Vertex(1),
+        v1 = new Vertex(3),
+        v2 = new Vertex(2),
+        v3 = new Vertex(5),
+        v4 = new Vertex(4);
+    const edges: Vertex[][] = [[v0, v1], [v1, v2], [v2, v3], [v0, v3], [v2, v4], [v3, v4]];
```
justin-tse, line 96:
```suggestion
    const edges = [[v0, v1], [v1, v2], [v2, v3], [v0, v3], [v2, v4], [v3, v4]];
```

codes/typescript/chapter_graph/graph_adjacency_list.ts
```diff
@@ -79,48 +76,51 @@ class GraphAdjList {
...
+    console.log("\n删除边 1-3 后，图为");
+    graph.print();
 
-/* 添加顶点 */
-const v5 = new Vertex(6);
-graph.addVertex(v5);
-console.log("\n添加顶点 6 后，图为");
-graph.print();
+    /* 添加顶点 */
+    const v5: Vertex = new Vertex(6);
```
justin-tse, line 114:
```suggestion
    const v5 = new Vertex(6);
```

codes/typescript/chapter_graph/graph_adjacency_list.ts
```diff
@@ -79,48 +76,51 @@ class GraphAdjList {
...
+if (require.main === module) {
+    /* Driver Code */
+    /* 初始化无向图 */
+    const v0 = new Vertex(1),
+        v1 = new Vertex(3),
+        v2 = new Vertex(2),
+        v3 = new Vertex(5),
+        v4 = new Vertex(4);
+    const edges: Vertex[][] = [[v0, v1], [v1, v2], [v2, v3], [v0, v3], [v2, v4], [v3, v4]];
+    const graph: GraphAdjList = new GraphAdjList(edges);
```
justin-tse, line 97:
```suggestion
    const graph = new GraphAdjList(edges);
```

justin-tse (approved):
Thank you!@zhuoqinyue Clear codes! My suggestions to use explicit types or implicit types are:
It's generally a good practice to use explicit types for function parameters and return types, as well as for variables that can be of more than one type. However, for variables that can only have one type and where the type is clear from the value, implicit typing is often sufficient.

zhuoqinyue:
OK, I've updated it

justin-tse (approved):
Thank you!@zhuoqinyue

krahets (approved):

PR #376, opened 2023-02-21
Update array.js

Rabbit937:
修改insert函数丢失最后一位元素


zhuoqinyue:
image
Your codes will cause a another problem.
Maybe you can use `let i = nums.length` instead of `let i = nums.length - 1`.

Rabbit937:
yes，you are right

krahets:
We can't avoid it because the length of an Array is fixed, which means when we insert an element, the last element of the original array must be lost.

PR #431, opened 2023-03-20
feat(counting_sort): support counting_sort in c/go

Reanon:

### DIFF
--- a/codes/c/chapter_sorting/counting_sort.c
+++ b/codes/c/chapter_sorting/counting_sort.c
@@ -1,0 +1,81 @@
+/**
+ * File: counting_sort.c
+ * Created Time: 2023-03-20
+ * Author: Reanon (793584285@qq.com)
+ */
+#include "../include/include.h"
+
+/* 计数排序 */
+// 简单实现，无法用于排序对象
+void countingSortNaive(int nums[], int size) {
+    // 1. 统计数组最大元素 m
+    int m = 0;
+    for (int i = 0; i < size; i++) {
+        if (nums[i] > m) {
+            m = nums[i];
+        }
+    }
+    // 2. 统计各数字的出现次数
+    // counter[num] 代表 num 的出现次数
+    int *counter = malloc(sizeof(int) * size);
+    for (int i = 0; i < size; i++) {
+        counter[nums[i]]++;
+    }
+    // 3. 遍历 counter ，将各元素填入原数组 nums
+    int i = 0;
+    for (int num = 0; num < m + 1; num++) {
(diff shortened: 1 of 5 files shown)


codes/c/chapter_sorting/counting_sort.c
```diff
@@ -0,0 +1,81 @@
...
+    // 1. 统计数组最大元素 m
+    int m = 0;
+    for (int i = 0; i < size; i++) {
+        if (nums[i] > m) {
+            m = nums[i];
+        }
+    }
+    // 2. 统计各数字的出现次数
+    // counter[num] 代表 num 的出现次数
+    int *counter = malloc(sizeof(int) * size);
```
Gonglja, line 20: Here the return value of malloc is `void *`, which needs to be converted to `int *`.

`int *counter =(int *) malloc(sizeof(int) * size);`

codes/c/chapter_sorting/counting_sort.c
```diff
@@ -0,0 +1,81 @@
...
```
Gonglja, line 46: the same as above

codes/c/chapter_sorting/counting_sort.c
```diff
@@ -0,0 +1,81 @@
...
+        counter[nums[i]]++;
+    }
+    // 3. 求 counter 的前缀和，将“出现次数”转换为“尾索引”
+    // 即 counter[num]-1 是 num 在 res 中最后一次出现的索引
+    for (int i = 0; i < m; i++) {
+        counter[i + 1] += counter[i];
+    }
+    // 4. 倒序遍历 nums ，将各元素填入结果数组 res
+    // 初始化数组 res 用于记录结果
+    int *res = malloc(sizeof(int) * size);
```
Gonglja, line 57: the same as above

codes/c/chapter_sorting/counting_sort.c
Reanon: It is unnecessary, as void * is automatically and safely promoted to any other pointer type in C.more in https://stackoverflow.com/questions/605845/do-i-cast-the-result-of-malloc

Gonglja: > It is unnecessary, as void * is automatically and safely promoted to any other pointer type in C.more in https://stackoverflow.com/questions/605845/do-i-cast-the-result-of-malloc

ok, thanks @Reanon 

I took a look, there is no problem in C (sorry, I just used the g++ compiler)

Reanon: 🤝

PR #450, opened 2023-04-06
✨ feat(rust/tree): add binary_tree_bfs, binary_tree_dfs, 

xBLACKICEx:

image
image

### DIFF
--- a/codes/rust/include/tree_node.rs
+++ b/codes/rust/include/tree_node.rs
@@ -5,6 +5,7 @@
  */
 
 use std::cell::RefCell;
+use std::collections::VecDeque;
 use std::rc::Rc;
 
 #[allow(dead_code)]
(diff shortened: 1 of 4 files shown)


krahets:
To be reviewed by @sjinzh

sjinzh (approved):
@krahets It's OK.

codes/rust/include/tree_node.rs
```diff
@@ -26,4 +27,65 @@ impl TreeNode {
...
+    let mut que = VecDeque::new();
+    que.push_back(Rc::clone(&root));
+
+    let mut index = 0;
+    while let Some(node) = que.pop_front() {
+        index += 1;
+        if index >= list.len() {
+            break;
+        }
+        if index < list.len() {
```
krahets, line 76: Incorrect. Please refer to the other language implementation

codes/rust/include/tree_node.rs
```diff
@@ -26,4 +27,65 @@ impl TreeNode {
...
+            node.borrow_mut().left = Some(TreeNode::new(list[index]));
+            que.push_back(Rc::clone(&node.borrow().left.as_ref().unwrap()));
+        }
+
+        index += 1;
+        if index >= list.len() {
```
krahets, line 82: Same as above

codes/rust/include/tree_node.rs
xBLACKICEx: Sorry, I can't find out what you said is Incorrect，from other languages. are you mean changing list.len () to variable n?

krahets: https://github.com/krahets/hello-algo/blob/2351fd1a1a0d73410ac3537df79d293bf2e205e9/codes/python/modules/binary_tree.py#L33

The function is designed to convert the list to any trees. So we should parse the null node from the list.
This line determines whether the node is null so we can handle the list like `[1, None, 2, 3, ...]`.

xBLACKICEx:
It should be ok by now,

xBLACKICEx:
It should be ok by now, i try some test output is same as python.

krahets (approved):
Thank @xBLACKICEx for the PR!

Issue #222, opened 2023-01-08
Code Transpilation to Go (Help Wanted)


Reanon:
We need your help translating the code to Go.

Please check the contribution guidelines at [#15](https://github.com/krahets/hello-algo/issues/15)

## TODO List

- [x] Computational Complexity
- - [x] Time Complexity https://github.com/krahets/hello-algo/pull/99 (@msk397)
- - [x] Space Complexity https://github.com/krahets/hello-algo/pull/119 (@Cathay-Chen)
- - [x] Space Time Tradeoff https://github.com/krahets/hello-algo/pull/102 (@Cathay-Chen) 
- - [x] Leetcode Two Sum https://github.com/krahets/hello-algo/pull/8 (@Reanon)
- [x] Array And LinkedList
- -  [x] Array https://github.com/krahets/hello-algo/pull/178 (@guowei-gong)
- -  [x] LinkedList https://github.com/krahets/hello-algo/pull/188 (@Cathay-Chen)
- -  [x] List https://github.com/krahets/hello-algo/pull/128 (@msk397)
-  [x] Stack And Queue
- -  [x] Stack https://github.com/krahets/hello-algo/pull/45 (@Reanon)
- -  [x] Queue https://github.com/krahets/hello-algo/pull/45 (@Reanon)
- -  [x] Deque https://github.com/krahets/hello-algo/pull/45 (@Reanon)
-  [x] Hashing
- -  [x] Hash Map https://github.com/krahets/hello-algo/pull/110 (@msk397)
- -  [ ] Hash Collision
-  [x] Tree
- -  [x] Binary Tree https://github.com/krahets/hello-algo/pull/20 (@Reanon)
- -  [x] Binary Tree Types https://github.com/krahets/hello-algo/pull/29 (@Reanon)
- -  [x] Binary Search Tree https://github.com/krahets/hello-algo/pull/29 (@Reanon)
- -  [x] AVL Tree https://github.com/krahets/hello-algo/pull/224 (@Reanon)
-  [x] Heap
- -  [x] Max Heap https://github.com/krahets/hello-algo/pull/246 (@Reanon)
-  [x] Searching
- -  [x] Linear Search https://github.com/krahets/hello-algo/pull/20 (@Reanon)
- -  [x] Binary Search https://github.com/krahets/hello-algo/pull/76 (@Slone123c )
- -  [x] Hash Search https://github.com/krahets/hello-algo/pull/96 (@Slone123c)
-  [x] Sorting
[...]

PR #505, opened 2023-05-18
feat(hashing/hash_map_collision_with_linkedlist)

shanghai-Jerry:
…tem when hash key collision in one bucket

Based on implemention  of arrayHashMap
add go code:  using linked list to store items when hash collision happens in one bucket.  CheckIssue here:  #222 
@Reanon @krahets  PTAL


krahets (commented):
Thanks for the clean implementation. But It is not in the outline of this book.

shanghai-Jerry:
> Thanks for the clean implementation. But It is not in the outline of this book.

？ it's the implementation about hash collision with linked list，why is outline of this book

krahets:
Hi @shanghai-Jerry , thanks for your attention!

It is because there was no corresponding position in the article for this implementation. We should gurantte every executable code block can be found both in the source code file and articles.

PR #507, opened 2023-05-19
Update hash_map.md

SeekingSofter:
The time complexity of inserting an element into an array is O(n).


krahets:
When using an array as a "map", we can easily add the element to the end of the array. So it is O(1).

PR #512, opened 2023-05-22
Add Q&A of computational_complexity and graph chapter

longsizhuo123:
Add the Q&A part about the computational complexity and graph chapter. And no code changes.


longsizhuo123:
Some question is too simple like "链接矩阵和图是怎么相关转换的". So I don't know if it had any necessary to add these to the Q&A. In my opinion, the answer may be more suitable in the main body of the chapter. So I didn't add this question to the Q&A part.

krahets:
Thanks for the PR! I agree that some questions are not suitable for the Q&A part, such as the one you mentioned.

However, I suggest collecting the questions with loose requirements, because we can discuss removing the inappropriate questions in the PR.

krahets (approved):
Thanks for the PR! I've made some changes. Please check them.

PR #554, opened 2023-06-14
feat(csharp): add top_k

hpstory:

Test Result:
topk


hpstory:
When I encountered some problems while editing the chapter on hashing, do you have any plans extracting the Pair class and replacing it with a struct?

krahets:
Hi @hpstory , thanks for the PR!

I'm considering extracting the Pair Class to modules because all the hash map class use it.

Yes. You can use struct if C# support it. Some languages like Java and Python do not support struct.

krahets (approved):

PR #568, opened 2023-06-26
feat(csharp/hashing): add code and update docs to chapter hashing

hpstory:

Test result:

### DIFF
--- a/codes/csharp/chapter_hashing/hash_map_chaining.cs
+++ b/codes/csharp/chapter_hashing/hash_map_chaining.cs
@@ -1,0 +1,146 @@
+﻿/**
+* File: hash_map_chaining.cs
+* Created Time: 2023-06-26
+* Author: hpstory (hpstory1024@163.com)
+*/
+
+namespace hello_algo.chapter_hashing;
+
+/* 链式地址哈希表 */
+class HashMapChaining {
+    int size; // 键值对数量
+    int capacity; // 哈希表容量
+    double loadThres; // 触发扩容的负载因子阈值
+    int extendRatio; // 扩容倍数
+    List<LinkedList<Pair>> buckets; // 桶数组
+
+    /* 构造方法 */
+    public HashMapChaining() {
+        size = 0;
+        capacity = 4;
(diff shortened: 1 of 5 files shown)


codes/csharp/chapter_hashing/hash_map_chaining.cs
```diff
@@ -0,0 +1,146 @@
...
+        buckets[index].AddLast(new Pair(key, val));
+        size++;
+    }
+
+    /* 删除操作 */
+    public void remove(int key) {
+        int index = hashFunc(key);
+        // 遍历桶，从中删除键值对
+        LinkedListNode<Pair> head = buckets[index].First;
+        while (head != null) {
```
hpstory, line 76: I use linked lists to implement this part. This is some different from the implementation of other languages.

hpstory: Add I didn't add Pair class to hash_map_chaining and hash_map_open_addressing, because in C# cannot allowed classes with the same name under a single namespace. I'm not sure if the docs needs to be updated. If you have any suggestions, please leave a message and I will update PR.

krahets: > I use linked lists to implement this part. This is some different from the implementation of other languages.

Better to use dynamic array for the consistency.

krahets: > And I didn't add Pair class to hash_map_chaining and hash_map_open_addressing, because in C# cannot allowed classes with the same name under a single namespace. I'm not sure if the docs needs to be updated. If you have any suggestions, please leave a message and I will update PR.

Thanks for pointing out the problem. I was thinking make the `Pair` class as a public module. No need more action and I will handle it.

hpstory: > > I use linked lists to implement this part. This is some different from the implementation of other languages.
> 
> Better to use dynamic array for the consistency.

fixed

krahets (approved):
Thanks fort the PR!

PR #592, opened 2023-07-10
fix(csharp): fix csharp method name in knapsack

hpstory:

1. fix method name in knapsack
2. add csharp code for min path sum
test result:

### DIFF
--- a/codes/csharp/chapter_dynamic_programming/min_path_sum.cs
+++ b/codes/csharp/chapter_dynamic_programming/min_path_sum.cs
@@ -9,14 +9,13 @@ namespace hello_algo.chapter_dynamic_programming;
 public class min_path_sum {
     /* 最小路径和：暴力搜索 */
     public int minPathSumDFS(int[][] grid, int i, int j) {
-        int INF = 0x3f3f3f3f;
         // 若为左上角单元格，则终止搜索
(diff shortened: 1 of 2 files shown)


codes/csharp/chapter_dynamic_programming/min_path_sum.cs
```diff
@@ -0,0 +1,126 @@
...
+        // 计算从左上角到 (i-1, j) 和 (i, j-1) 的最小路径代价
+        int left = minPathSumDFS(grid, i - 1, j);
+        int up = minPathSumDFS(grid, i, j - 1);
+        // 返回从左上角到 (i, j) 的最小路径代价
+        return Math.Min(left, up) + grid[i][j];
+    }
+
+    /* 最小路径和：记忆化搜索 */
+    public int minPathSumDFSMem(int[][] grid, int[][] mem, int i, int j) {
+        int INF = 0x3f3f3f3f;
```
hpstory, line 30: I don’t know how to modify this. In C#, int.MaxValue should not be used directly, as there may be an overflow in the following addition. Usually, 0x3f is used instead. Should I use int.MaxValue directly or explain why 0x3f is used?

krahets (approved):
Thanks for the quick fix!

PR #632, opened 2023-07-20
feat(csharp): add csharp code for array binary tree

hpstory:

1. add code for array binary tree
2. rewrite utils/TreeNode.cs
3. add extension method for print nullable list 

Test result:

### DIFF
--- a/codes/csharp/chapter_tree/array_binary_tree.cs
+++ b/codes/csharp/chapter_tree/array_binary_tree.cs
@@ -1,0 +1,134 @@
+﻿/**
+* File: array_binary_tree.cs
+* Created Time: 2023-07-20
+* Author: hpstory (hpstory1024@163.com)
+*/
+
+namespace hello_algo.chapter_tree;
+
+/* 数组表示下的二叉树类 */
+public class ArrayBinaryTree {
+    private List<int?> tree;
+
+    /* 构造方法 */
+    public ArrayBinaryTree(List<int?> arr) {
+        tree = new List<int?>(arr);
+    }
+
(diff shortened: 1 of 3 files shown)


codes/csharp/chapter_tree/array_binary_tree.cs
```diff
@@ -0,0 +1,134 @@
...
+
+    /* 后序遍历 */
+    public List<int> postOrder() {
+        List<int> res = new List<int>();
+        dfs(0, "post", res);
+        return res;
+    }
+}
+
+public class array_binary_tree {
```
hpstory, line 97: This part of code is different with others. Because ArrayBinaryTree class doesn't have a default constructor for test framework, resulting in a test failure, so move it outside of the class.

krahets: I see. C#'s rule is the same with Java.

krahets (approved):
Thanks for the clean code!

codes/csharp/utils/TreeNode.cs
```diff
@@ -13,50 +13,59 @@ public class TreeNode {
...
-                queue.Enqueue(node.right);
-            } else {
-                list.Add(null);
-            }
+    /* 将列表反序列化为二叉树 */
+    public static TreeNode? ListToTree(List<int?> arr) {
+        return ListToTreeDFS(arr, 0);
+    }
+
+    /* 将二叉树序列化为列表：递归 */
```
chengxuyuanluyu, line 53: nb

Issue #479, opened 2023-04-30
rust "binary_search" return type -> choose uszie or i32 ?


XIAOZHUXUEJAVA:
I think the binary search in Rust could be written like this:
```rust
pub fn binary_search(nums: &[i32], target: i32) -> Option<usize> {
        let mut low = 0;
        let mut high = nums.len();
        while low < high {
            let mid = (high - low) / 2 + low;
            match target.cmp(&nums[mid]) {
                std::cmp::Ordering::Equal => return Some(mid),
                std::cmp::Ordering::Less => high = mid,
                std::cmp::Ordering::Greater => low = mid + 1,
            }
        }
        None
}
```
This is more in line with Rust's language features， maybe

krahets:
@sjinzh @xBLACKICEx

krahets:
Simplicity is one of the principles of this book. 

This looks like a language syntax sugar that doesn't suppress the `if ... else ... ` in performance or simplicity.

```
            match target.cmp(&nums[mid]) {
                std::cmp::Ordering::Equal => return Some(mid),
                std::cmp::Ordering::Less => high = mid,
                std::cmp::Ordering::Greater => low = mid + 1,
            }
```

Using `usize` is more formal and with greater adaptivity. However, it is not vital for the subject of data structure and algorithms.

XIAOZHUXUEJAVA:
The code use of `while i <= j`  and `j = m - 1` in the code may lead to boundary issues (left inclusive, right inclusive interval). 

In Rust, array indices must be of non-negative integer type `usize`. This means that if we want to access an element of an array, we must provide a non-negative integer index, **which means that the code will fail when the target value is smaller than all array elements.**

Of course, my programming level is not high, and my analysis may not be accurate.

Of course, you are not wrong that data structures and algorithms are indeed largely language-independent, but in my view, there is a problem with this code. If you do not think so, you can ignore this.

```rust
    #[test]
    fn test_binary_search() {
        let nums = [2, 3, 4, 7, 9];
        assert_eq!(-1, BinarySearch::binary_search_i(&nums, 1));
    }
```
For instance:
![image](https://user-images.githubusercontent.com/81864199/235947670-1d1f78e8-2848-41f8-af97-a294e9480d2c.png)
![image](https://user-images.githubusercontent.com/81864199/235947852-5c7bd68c-015e-4a09-869a-896cbcea6d5d.png)

krahets:
[...]

PR #685, opened 2023-08-13
Fix the bug of index in rust codes

night-cruise:
See 👉 https://github.com/krahets/hello-algo/discussions/52?sort=new#discussioncomment-6708590


krahets (approved):
Thanks for the fix!

See #479.

PR #722, opened 2023-09-05
FixedBug: Update linkedlist_queue.java

lyl625760:
pop后，如果队列已经空，尾元素设置为null，这样最后一个节点的资源才能释放掉。 再次push时状态是正确的。


lyl625760:
If it's true, we don't have to release anything because  the programm  is running ..... 
the behavior is not consistent  when the rear is pointed to an unused node .

krahets:
> the behavior is not consistent when the rear is pointed to an unused node .

`queSize` is used to determine whether the queue is empty, not `front` and `rear`. The queue operations are independent of the value of `rear`.

PR #756, opened 2023-09-13
Remove duplication in cmake

Nigh:
Removed duplicate targets in `CMakeLists` that cause CMake errors.


krahets (approved):
Thanks for the fix!

PR #773, opened 2023-09-20
feat: add unit test for linearSearchArray method

ahnguyentran:
This PR is about adding a unit test for the linearSearchArray method. It is a requirement for my exam, sorry if it borders but if it helps I'm very happy.
Thank you!


krahets:
Thanks for the PR! We only provide a basic test case and leave the rest to the reader's discretion.

PR #812, opened 2023-09-30
Update suggestions

Evilrabbit520:
Here I think using "code warehouse" instead of "code warehouse" can eliminate unnecessary ambiguity.


krahets (approved):
Thanks for the PR!

PR #823, opened 2023-10-02
fix: Removed unnecessary JS codes in linked_list.md

logan-qiu:

### DIFF
--- a/docs/chapter_array_and_linkedlist/linked_list.md
+++ b/docs/chapter_array_and_linkedlist/linked_list.md
@@ -96,8 +96,6 @@
     ```javascript title=""
     /* 链表节点类 */
     class ListNode {
-        val;
-        next;
         constructor(val, next) {
             this.val = (val === undefined ? 0 : val);       // 节点值
             this.next = (next === undefined ? null : next); // 指向下一节点的引用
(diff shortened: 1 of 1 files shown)


logan-qiu:
@krahets k神，我认为实际操作中JS只需要以这样的写法就可以了：
image
按照修改后的理解可能会更加直观

justin-tse (commented):
@logan-qiu Thank you! I agree that removing it will be better for JS in this place. But this is also a correct way to define a class, making it more documented. This is called field declarations. You can see some reference from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes)
image

krahets:
LeetCode also uses the simplified definition. Do you think it is better to merge this PR @justin-tse ?

```
/**
 * Definition for singly-linked list.
 * function ListNode(val, next) {
 *     this.val = (val===undefined ? 0 : val)
 *     this.next = (next===undefined ? null : next)
 * }
 */
```

logan-qiu:
@justin-tse Hey Justin, thank you for the reference!

I understand the original purpose to declare fields up-front would help with optimization, but I believe this is not a common approach in a real-world scenario. Also it is slightly less readable.

image

Maybe we can do an optional page with introducing private class features? Just an idea, we can definitely go with the existing one but personally I like it to be simpler and more readable.

justin-tse:
@logan-qiu @krahets Thank you! I also prefer the simplified definition.

justin-tse:
@logan-qiu Could you help to update the Javascript code for this not-common style?

logan-qiu:
@justin-tse May I also make adjustments with other code blocks with field declarations and re-submit the PR?

justin-tse:
@logan-qiu Thank you, please do that.

krahets:
@logan-qiu It is more convenient to add commits to your forked repo, which will be synchronized to this PR automatically.

PR #836, opened 2023-10-03
Update hash_algorithm.md

theNefelibatas:
- Add the hiding property required for the hash algorithm in security domain.


krahets (approved):
Thanks for the PR!

PR #839, opened 2023-10-07
fix: Fixed the state transition error in minCostClimbingStairsDP and minCostClimbingStairsDPComp function

foursevenlove:
- This pull request addresses an issue and fixes it by correctly updating the state transition in the `minCostClimbingStairsDP` and `minCostClimbingStairsDPComp` function.

- In the `minCostClimbingStairsDP` function, the previous implementation had an error in the state transition logic where it was incorrectly choosing to `int(math.Min(float64(dp[i-1]), float64(dp[i-2]+cost[i])))` instead of `int(math.Min(float64(dp[i-1]+cost[i]), float64(dp[i-2]+cost[i])))`. In order to improve readability, I have adopted the following approach and implemented the `min` function: `min(dp[i-1], dp[i-2]) + cost[i]`.

- For the `minCostClimbingStairsDPComp` function, the modification is the same as above.

### DIFF
--- a/codes/go/chapter_dynamic_programming/min_cost_climbing_stairs_dp.go
+++ b/codes/go/chapter_dynamic_programming/min_cost_climbing_stairs_dp.go
@@ -4,13 +4,17 @@
 
 package chapter_dynamic_programming
 
-import "math"
-
 /* 爬楼梯最小代价：动态规划 */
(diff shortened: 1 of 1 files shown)


Reanon (approved):

krahets (approved):
Thanks for the PR!

PR #842, opened 2023-10-08
docs: add formula derivation for unbounded knapsack

hpstory:


krahets:
Thanks for the PR! It is mathmatically correct, but I think it makes the problem more complicated unneccessarily.

Assume that we're calculating $dp[i, c]$ using $dp[i, c - wgt[i]]$, according to the state definition and the state transfer sequence, we can find:

- $dp[i, c - wgt[i]]$ has included considering the $dp[i, c - 2 \cdot wgt[i]]$.
- $dp[i, c - 2 \cdot wgt[i]]$ has considered $dp[i, c - 3 \cdot wgt[i]]$.
- ...

That is, $dp[i, c - wgt[i]]$ has accumulated the greatest result in the past. We can only use it to calculate $dp[i, c]$.

PR #870, opened 2023-10-19
Update array_hash_map.c

amlei:
完善哈希表键值对获取

- [x] I've tested the code and ensured the outputs are the same as the outputs of reference code.
- [x] I've checked the code (formatting, comments, indentation, file header, etc) carefully.

### DIFF
--- a/codes/c/chapter_hashing/array_hash_map.c
+++ b/codes/c/chapter_hashing/array_hash_map.c
@@ -27,6 +27,11 @@ typedef struct {
     int len;
 }IndexMembers;
 
+/* 基于数组简易实现的哈希表 */
+typedef struct {
+    Pair *buckets[HASH_MAP_DEFAULT_SIZE];
+} ArrayHashMap;
+
 /* 初始化存储键 */
 void InitIndex(IndexMembers *index){
     index->len = -1;
@@ -37,10 +42,10 @@ void putKey(IndexMembers *index, int key){
     index->key[++index->len] = key;
 }
 
-/* 基于数组简易实现的哈希表 */
-typedef struct {
-    Pair *buckets[HASH_MAP_DEFAULT_SIZE];
-} ArrayHashMap;
+/* 删除哈希键 */
+void DelKey(IndexMembers *index){
+    --index->len;
+}
 
 /* 哈希表初始化函数 */
 ArrayHashMap *newArrayHashMap() {
@@ -77,11 +82,11 @@ void put(ArrayHashMap *d,IndexMembers *indexMembers, const int key, const char *
 }
 
 /* 删除操作 */
-void removeItem(ArrayHashMap *d, const int key) {
+void removeItem(ArrayHashMap *d, IndexMembers *indexMembers, const int key) {
     int index = hashFunc(key);
+
     free(d->buckets[index]->val);
-    free(d->buckets[index]);
-    d->buckets[index] = NULL;
+    DelKey(indexMembers);
 }
 
 /* 获取所有键值对 */
@@ -89,7 +94,6 @@ void pairSet(ArrayHashMap *d, IndexMembers indexMembers, MapSet *set) {
     Pair *entries;
     int i = 0, index = 0;
     int total = indexMembers.len;
-
 
     printf("有效键值对数量:%d\n", total + 1);
 
@@ -140,15 +144,17 @@ void valueSet(ArrayHashMap *d, IndexMembers indexMembers, MapSet *set) {
 /* 打印哈希表 */
 void print(ArrayHashMap *d, IndexMembers indexMembers) {
     int i;
-    MapSet set;
-    pairSet(d, indexMembers, &set);
-    Pair *entries = (Pair *)set.set;
+    MapSet *set;
+    set = malloc(indexMembers.len);
+    pairSet(d, indexMembers, set);
 
-    for (i = 0; i <= set.len; i++) {
+    Pair *entries = (Pair *)set->set;
+
+    for (i = 0; i <= set->len; i++) {
         printf("%d -> %s\n", entries[i].key, entries[i].val);
     }
 
-    /* free(set.set); */
+    free(set);
 }
 
 /* Driver Code */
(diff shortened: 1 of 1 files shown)


krahets:
Hi @amlei , please make changes based on the latest code. I found that this PR's code is based on a previous depreciated code.

krahets:
Pending for the review by @Gonglja

codes/c/chapter_hashing/array_hash_map.c
```diff
@@ -21,11 +21,32 @@ typedef struct {
     int len;
 } MapSet;
 
+/* 用以索引键获取数据 */
```
Gonglja, line 24: 用于？

codes/c/chapter_hashing/array_hash_map.c
```diff
@@ -21,11 +21,32 @@ typedef struct {
...
+typedef struct {
+    int key[HASH_MAP_DEFAULT_SIZE];
+    int len;
+}IndexMembers;
```
Gonglja, line 28: `IndexMembers` appears from the code to be part of the `ArrayHashMap` structure, and would be more appropriately placed in an `ArrayHashMap`.

Subsequent function parameters need not be passed.
`IndexMembers *indexMembers,`

codes/c/chapter_hashing/array_hash_map.c
```diff
@@ -48,162 +69,145 @@ const char *get(const ArrayHashMap *d, const int key) {
...
     char **vals = (char **)set.set;
     printf("\n单独遍历键 Value\n");
     for (i = 0; i < set.len; i++) {
         printf("%s\n", vals[i]);
     }
-    free(set.set);
 
     return 0;
 }
+
```
Gonglja, line 213: Lack of a free function, memory space is created by `newArrayHashMap` but not freed.

codes/c/chapter_hashing/array_hash_map.c
amlei: ``MapSet``  由 @Guanngxu  编写，运用于遍历输出时的数据集

codes/c/chapter_hashing/array_hash_map.c
amlei: Yes, but I found an exception while obtaining the ``KeySet`` and ``ValueSet``. Use ``IndexMembers *indexMembers`` is to ensure the right key address.

codes/c/chapter_hashing/array_hash_map.c
amlei: Ok, I added now.

codes/c/chapter_hashing/array_hash_map.c
krahets: /* 用以索引键获取数据 */  -> /* 用于索引键获取数据 */

krahets:
Hi @amlei , could you explain the problem this PR try to solve?

codes/c/chapter_hashing/array_hash_map.c
```diff
@@ -21,11 +21,32 @@ typedef struct {
...
+}
+
+/* 将哈希键存入数组 */
+void putKey(IndexMembers *index, int key){
+    index->key[++index->len] = key;
+}
+
+/* 删除哈希键 */
+void DelKey(IndexMembers *index){
+    --index->len;
```
krahets, line 47: How do you guarantte deleting the target key from `*index`

krahets: Assuming that we call `removeItem(..., key1)` twice, this function will delete two keys(len decreasing by 2), but actually there is just a single target key `key1`.

codes/c/chapter_hashing/array_hash_map.c
krahets: @amlei Could you explain it in detail?

amlei:
> Hi @amlei , could you explain the problem this PR try to solve?

Solve the memory overflow caused when obtaining key-value pairs and cannot implement an array-based hash table.

krahets:
> > Hi @amlei , could you explain the problem this PR try to solve?
> 
> Solve the memory overflow caused when obtaining key-value pairs and cannot implement an array-based hash table.

Could you provide a test case?

amlei:
> > > Hi @amlei , could you explain the problem this PR try to solve?
> > 
> > 
> > Solve the memory overflow caused when obtaining key-value pairs and cannot implement an array-based hash table.
> 
> Could you provide a test case?


In previous original code:
```C
int index = 0;
for (i = 0; i < HASH_MAP_DEFAULT_SIZE; i++) {
    keys[index] = d->buckets[i]->key;
    index++;
}

```

But key address is ``key % HASH_MAP_DEFAULT_SIZE`` that if use the code will be memory overflow. Because, the index ranges from 0 to a value less than the effective amount of data, but the data stored for key value pairs is not stored in increments of 0. Rather based on ``hashFunc()`` return ``key % HASH_MAP_DEFAULT_SIZE``, so I try use list stored the data of keys. This can solve the problem of outputting all key value pairs. Of course, it is base on ``array_hash_map`` feature.

Code example:
```C
Pair *cur = map->buckets[12836 % HASH_MAP_DEFAULT_SIZE];
printf("%d -> %s\n", map->buckets[0]->key,map->buckets[0]->val);
printf("%d -> %s\n",cur->key, cur->val);
```
Output:
```C
-1498513520 -> P��)
12836 -> 小哈
```

krahets:
> Because, the index ranges from 0 to a value less than the effective amount of data, but the data stored for key value pairs is not stored in increments of 0.

This is how hash table works. The keys are distributed in the buckets after mapping with hash function.

> Pair *cur = map->buckets[12836 % HASH_MAP_DEFAULT_SIZE];
> printf("%d -> %s\n", map->buckets[0]->key,map->buckets[0]->val);
> printf("%d -> %s\n",cur->key, cur->val);

We should find corresponding value of a key using `get()` function rather than access `map->buckets`. The buckets is private to the users. When calling `get(0)`, it will return `NULL` which represent the key does not exists.

krahets:
I think `IndexMembers` is for:

1. Continuously store the key to make it easier to traverse the hashmap.
2. `len` is used to Indicate the size of the hashmap.

Do I miss something?

amlei:
Yes,yes!

krahets:
Thanks for the answer!

This PR won't be merged for these reasons:

1. Adding `IndexMembers` is not related to the memory overflow. Hash table naturally store key-val pairs based on hash function mapping.
2. Storing keys in a list continuously could be utilized to optimize traversing the hash table. Currently this is not in the outline of the book.

By the way, if we add a lists to optimize traversal, we should maintain its operations correctly(add and remove keys from the list). Removing a key from a list should take $O(n)$ time. But I believe there exists a more elegant solution.

PR #905, opened 2023-10-29
Add missing terminologies

yuelinxin:
Added two missing terminologies that are present in the book:
* AVL tree
* Red Black tree


krahets (approved):
Thanks, @yuelinxin !

PR #915, opened 2023-11-03
Update terminology.md

JackYang-hellobobo:
对应链表linked list 为 linked list node ，而不应该是对应下文的列表list node 。


krahets (approved):
Thanks for the fix! The term "list node" is commonly understood to refer to a node in a linked list, especially when the context has already made it clear that the discussion is about linked lists. However, without additional context, the term "list node" might be somewhat ambiguous.

PR #916, opened 2023-11-03
[TypeScript] fix: deep copy old hash table when resizing

eric183:
The bucketsTmp variable only holds a reference to the original hash table,  which causes all key-value pairs to be lost after resizing.

Fix this by deep copying the old hash table before resizing,  using Array.prototype.slice to copy array by value.

This ensures bucketsTmp has a full copy of the old hash table, and modifications to the new hash table will not affect it.

### DIFF
--- a/codes/typescript/chapter_hashing/hash_map_chaining.ts
+++ b/codes/typescript/chapter_hashing/hash_map_chaining.ts
@@ -93,7 +93,7 @@ class HashMapChaining {
     /* 扩容哈希表 */
     #extend(): void {
         // 暂存原哈希表
-        const bucketsTmp = this.#buckets;
+        const bucketsTmp = [...this.#buckets];
         // 初始化扩容后的新哈希表
         this.#capacity *= this.#extendRatio;
         this.#buckets = new Array(this.#capacity).fill(null).map((x) => []);


justin-tse (commented):
@eric183 Thank you, it is a better practice, you are creating a shallow copy of the original #buckets array using the spread operator, which can help prevent potential side effects from modifying the original buckets during the rehashing process.

krahets:
Hi @eric183, in extending operation we initialize a new buckets, and copy all key-val pairs from the old buckets to the new one. After that, the old bucket is no longer needed.

> which causes all key-value pairs to be lost after resizing.

I don't think the pairs will be lost after resizing the hash table because they have been added to the new buckets. Thus I believe create a deep copy of the original buckets is not necessary.

justin-tse:
The original code does not have problems, because have copied it into bucketsTmp, and recreate the memory for the extended bucket.


> Hi @eric183, in extending operation we initialize a new buckets, and copy all key-val pairs from the old buckets to the new one. After that, the old bucket is no longer needed.
> 
> > which causes all key-value pairs to be lost after resizing.
> 
> I don't think the pairs will be lost after resizing the hash table because they have been added to the new buckets. Thus I believe it is not necessary to create a deep copy of the original buckets.

@krahets Thank you for correcting me. The original code does not have problems because it has been copied into bucketsTmp, so when we resize the bucket, we do not lose any data. Then we recreated the memory for the extended bucket. So we can close this PR.

PR #917, opened 2023-11-03
[JavaScript] fix: deep copy old hash table when resizing

eric183:
The bucketsTmp variable only holds a reference to the original hash table,  which causes all key-value pairs to be lost after resizing.

Fix this by deep copying the old hash table before resizing,  using Array.prototype.slice to copy array by value.

This ensures bucketsTmp has a full copy of the old hash table, and modifications to the new hash table will not affect it.

### DIFF
--- a/codes/javascript/chapter_hashing/hash_map_chaining.js
+++ b/codes/javascript/chapter_hashing/hash_map_chaining.js
@@ -91,7 +91,7 @@ class HashMapChaining {
     /* 扩容哈希表 */
     #extend() {
         // 暂存原哈希表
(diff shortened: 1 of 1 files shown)


justin-tse (commented):
@eric183 Thank you, it is a better practice, you are creating a shallow copy of the original #buckets array using the spread operator, which can help prevent potential side effects from modifying the original buckets during the rehashing process.

justin-tse:
Follow this comment
https://github.com/krahets/hello-algo/pull/916#issuecomment-1793768654

PR #929, opened 2023-11-08
feat: add csharp sdk to dockerfile

hpstory:

- [x] I've tested the code and ensured the outputs are the same as the outputs of reference code.
- [x] I've checked the code (formatting, comments, indentation, file header, etc) carefully.

### DIFF
--- a/codes/.env
+++ b/codes/.env
@@ -1,0 +1,1 @@
+LANGS_SDK="python cpp java csharp"
--- a/codes/.env
+++ b/codes/.env
@@ -1,1 +1,0 @@
-LANGS_SDK="python cpp java csharp"
--- a/codes/docker-compose.yml
+++ b/codes/docker-compose.yml
@@ -4,7 +4,9 @@ services:
     build:
       context: .
       args:
-        LANGS: ${LANGS_SDK}
+        # 设置需要安装的语言，使用空格隔开
+        # Set the languages to be installed, separated by spaces
+        LANGS: "python cpp csharp"
     image: hello-algo-code
     container_name: hello-algo-code
     stdin_open: true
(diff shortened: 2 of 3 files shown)


hpstory:
Use docker-compose --env-file .\codes\.env -f .\codes\docker-compose.yml up

codes/.env
```diff
@@ -0,0 +1 @@
+LANGS_SDK="python cpp java csharp"
```
krahets, line 1: The languages to be installed are selected using `LANGS` arg in Dockerfile. Is this duplicated?

codes/docker-compose.yml
```diff
@@ -1,7 +1,10 @@
 version: '3.8'
 services:
   hello-algo-code:
-    build: .
+    build:
+      context: .
+      args:
+        LANGS: ${LANGS_SDK}
```
krahets, line 7: Same as above.

codes/.env
hpstory: Actually not, the LANGS in the dockerfile entirely depend on the variables in the docker-compose file. I think we need a default value here, but the dockerfile still needs to update.

codes/docker-compose.yml
```diff
@@ -1,7 +1,12 @@
 version: '3.8'
 services:
   hello-algo-code:
-    build: .
+    build:
+      context: .
+      args:
+        # 设置需要安装的语言，使用空格隔开
+        # Set the languages to be installed, separated by spaces
+        LANGS: "python cpp csharp"
```
krahets, line 9: I prefer setting the LANGS here with out adding a .ENV file. How do you think of it?

codes/.env
krahets: OK, I removed the default values of `LANGS` in Dockerfile to make it fully controlled by `docker-compose`.

codes/docker-compose.yml
hpstory: remove env file

krahets (approved):
Thanks for the PR!

PR #931, opened 2023-11-09
Unsize type must be greater than or equal to 0

night-cruise:
See here: https://github.com/krahets/hello-algo/discussions/32#discussioncomment-7378599


krahets (approved):
Thanks for the fix!

PR #951, opened 2023-11-18
Fix the type annotation in Rust

night-cruise:
See here 👉 https://github.com/krahets/hello-algo/discussions/46#discussioncomment-7538370


krahets (approved):
Thanks for the fix!

PR #954, opened 2023-11-18
translation: Update terminology and improve readability in preface summary

longsizhuo:
This commit made a few adjustments in the 'summary.md' file for clearer and more accessible language. "Brushing tool library" was replaced with "Coding Toolkit" to better reflect common terminology. Also, advice for beginners in algorithm learning journey was reformulated to imply a more positive approach avoiding detours and common pitfalls. The section related to the discussion forum was rewritten to sound more inviting to readers.

### DIFF
--- a/docs-en/chapter_preface/summary.md
+++ b/docs-en/chapter_preface/summary.md
@@ -2,6 +2,7 @@
 
 - The main audience of this book is beginners in algorithm. If you already have some basic knowledge, this book can help you systematically review your algorithm knowledge, and the source code in this book can also be used as a "Coding Toolkit".
 - The book consists of three main sections, Complexity Analysis, Data Structures, and Algorithms, covering most of the topics in the field.
-- For newcomers to algorithms, it is crucial to read an introductory book in the beginning stages to avoid many detours or common pitfalls.- Animations and graphs within the book are usually used to introduce key points and difficult knowledge. These should be given more attention when reading the book.
+- For newcomers to algorithms, it is crucial to read an introductory book in the beginning stages to avoid many detours or common pitfalls.
+- Animations and graphs within the book are usually used to introduce key points and difficult knowledge. These should be given more attention when reading the book.
 - Practice is the best way to learn programming. It is highly recommended that you run the source code and type in the code yourself.
 - Each chapter in the web version of this book features a discussion forum, and you are welcome to share your questions and insights at any time.
--- a/docs-en/chapter_preface/summary.md
+++ b/docs-en/chapter_preface/summary.md
@@ -1,8 +1,7 @@
 # Summary
 
-- The main audience of this book is beginners in algorithm. If you already have some basic knowledge, this book can help you systematically review your algorithm knowledge, and the source code in this book can also be used as a "brushing tool library".
+- The main audience of this book is beginners in algorithm. If you already have some basic knowledge, this book can help you systematically review your algorithm knowledge, and the source code in this book can also be used as a "Coding Toolkit".
 - The book consists of three main sections, Complexity Analysis, Data Structures, and Algorithms, covering most of the topics in the field.
-- For newcomers to algorithms, it is crucial to read an introductory book in the beginning stages to take the road less traveled.
-- Animations and graphs within the book are usually used to introduce key points and difficult knowledge. These should be given more attention when reading the book.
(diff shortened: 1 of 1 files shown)


docs-en/chapter_preface/summary.md
```diff
@@ -1,8 +1,7 @@
 # Summary
 
-- The main audience of this book is beginners in algorithm. If you already have some basic knowledge, this book can help you systematically review your algorithm knowledge, and the source code in this book can also be used as a "brushing tool library".
+- The main audience of this book is beginners in algorithm. If you already have some basic knowledge, this book can help you systematically review your algorithm knowledge, and the source code in this book can also be used as a "Coding Toolkit".
```
longsizhuo, line 3: Brushing tool library is not right, it is not a good translation for "刷题"

docs-en/chapter_preface/summary.md
```diff
@@ -1,8 +1,7 @@
...
 - The book consists of three main sections, Complexity Analysis, Data Structures, and Algorithms, covering most of the topics in the field.
-- For newcomers to algorithms, it is crucial to read an introductory book in the beginning stages to take the road less traveled.
-- Animations and graphs within the book are usually used to introduce key points and difficult knowledge. These should be given more attention when reading the book.
```
longsizhuo, line 6: “to take the road less traveled”意味着选择不寻常或不常走的道路，这与原文的意思有所不同。原文强调的是通过阅读入门书籍可以避免常见错误，而不是走一条不常走的路。

"To take the road less traveled" means to take an unusual or less traveled road, which is different from the original meaning. The original article emphasizes that common mistakes can be avoided by reading introductory books, rather than taking a road less traveled.

longsizhuo: https://en.wikipedia.org/wiki/Wikipedia:Taking_the_road_less_traveled

docs-en/chapter_preface/summary.md
```diff
@@ -1,8 +1,7 @@
...
+- For newcomers to algorithms, it is crucial to read an introductory book in the beginning stages to avoid many detours or common pitfalls.- Animations and graphs within the book are usually used to introduce key points and difficult knowledge. These should be given more attention when reading the book.
 - Practice is the best way to learn programming. It is highly recommended that you run the source code and type in the code yourself.
```
longsizhuo, line 6: be more polite

yuelinxin, line 6: please correct the formats, there should be a line break before "-Animations"

longsizhuo: fix it, ty

docs-en/chapter_preface/summary.md
```diff
@@ -1,8 +1,8 @@
 # Summary
 
-- The main audience of this book is beginners in algorithm. If you already have some basic knowledge, this book can help you systematically review your algorithm knowledge, and the source code in this book can also be used as a "brushing tool library".
+- The main audience of this book is beginners in algorithm. If you already have some basic knowledge, this book can help you systematically review your algorithm knowledge, and the source code in this book can also be used as a "Coding Toolkit".
 - The book consists of three main sections, Complexity Analysis, Data Structures, and Algorithms, covering most of the topics in the field.
-- For newcomers to algorithms, it is crucial to read an introductory book in the beginning stages to take the road less traveled.
+- For newcomers to algorithms, it is crucial to read an introductory book in the beginning stages to avoid many detours or common pitfalls.
 - Animations and graphs within the book are usually used to introduce key points and difficult knowledge. These should be given more attention when reading the book.
```
yanedie, line 6: Chinese：书内的动画和图解通常用于介绍重点和难点知识。阅读本书时，应给予这些内容更多关注。
Animations and graphs should be given more attention as they are used to introduce key points and difficult knowledge.

English is an explicit and upfront language. I think it's better to place the key part of the sentence at the beginning. And since there is a context, I delete "within the book".

longsizhuo: 我也认为这个部分在这里语意不通顺，但我认为可以先按照原文翻译，尽量不改动，除非有错误。

I also believe that this part is not clear in the context of this sentence, but I think we should translate it as close to the original text as possible, unless there are errors.

yuelinxin (approved):
looks good to me! thank you @longsizhuo

yanedie (approved):

krahets (approved):
Thanks for the PR!

PR #996, opened 2023-12-22
Fix the issue of discuss comment 7919887

night-cruise:
See here 👉 https://github.com/krahets/hello-algo/discussions/27#discussioncomment-7919887


krahets (approved):
Thanks for the fix!

PR #1006, opened 2023-12-25
Fix the issue of the comment 7938404

night-cruise:
See here 👉 https://github.com/krahets/hello-algo/discussions/52#discussioncomment-7938404


krahets (approved):
Thanks for the fix!

PR #1012, opened 2023-12-27
translation: Update chapter_data_structure/number_encoding.md

YangXuanyi:

### DIFF
--- a/docs-en/chapter_data_structure/number_encoding.md
+++ b/docs-en/chapter_data_structure/number_encoding.md
@@ -2,11 +2,11 @@
 
 !!! note
 
-    In this book, chapters marked with an * symbol are optional reads. If you are short on time or find them challenging, you may skip these initially and return to them after completing the essential chapters.
+    In this book, chapters marked with an * symbol are optional to read. If you are short on time or find them challenging, you may skip these initially and return to them after completing the essential chapters.
 
 ## Integer Encoding
 
-In the table from the previous section, we noticed that all integer types can represent one more negative number than positive numbers, such as the `byte` range of $[-128, 127]$. This phenomenon, somewhat counterintuitive, is rooted in the concepts of sign-magnitude, one's complement, and two's complement encoding.
+In the table from the previous section, we noticed that all integer types can represent one more negative number than positive numbers, such as the `byte` range of $[-128, 127]$. This phenomenon seems counterintuitive, and its underlying reason involves knowledge of sign-magnitude, ones' complement, and two's complement.
 
 Firstly, it's important to note that **numbers are stored in computers using the two's complement form**. Before analyzing why this is the case, let's define these three encoding methods:
 
@@ -16,7 +16,7 @@ Firstly, it's important to note that **numbers are stored in computers using the
 
 The following diagram illustrates the conversions among sign-magnitude, one's complement, and two's complement:
 
-![Conversions between Sign-Magnitude, One's Complement, and Two's Complement](number_encoding.assets/1s_2s_complement.png)
+![Conversions between Sign-magnitude, One's Complement, and Two's Complement](number_encoding.assets/1s_2s_complement.png)
 
 Although sign-magnitude is the most intuitive, it has limitations. For one, **negative numbers in sign-magnitude cannot be directly used in calculations**. For example, in sign-magnitude, calculating $1 + (-2)$ results in $-3$, which is incorrect.
 
@@ -65,7 +65,7 @@ Adding $1$ to the one's complement of negative zero produces a carry, but with `
 
 One last puzzle is the $[-128, 127]$ range for `byte`, with an additional negative number, $-128$. We observe that for the interval $[-127, +127]$, all integers have corresponding sign-magnitude, one's complement, and two's complement, and these can be converted between each other.
 
-However, **the two's complement $1000 \; 0000$ is an exception without a corresponding sign-magnitude**. According to the conversion method, its sign-magnitude would be $0000 \; 0000$, which is a contradiction since this represents zero, and its two's complement should be itself. Computers designate this special two's complement $1000 \; 0000$ as representing $-128$. In fact, the calculation of $(-1) + (-127)$ in two's complement results in $-128$.
+However, **the two's complement $1000 \; 0000$ is an exception without a corresponding sign-magnitude**. According to the conversion method, its sign-magnitude would be $0000 \; 0000$, which is a contradiction because it represents zero, and its two's complement should be itself. Computers designate this special two's complement $1000 \; 0000$ as representing $-128$. In fact, the calculation of $(-1) + (-127)$ in two's complement results in $-128$.
 
 $$
 \begin{aligned}
--- a/docs-en/chapter_data_structure/number_encoding.md
+++ b/docs-en/chapter_data_structure/number_encoding.md
@@ -2,11 +2,11 @@
 
 !!! note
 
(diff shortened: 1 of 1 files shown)


yuelinxin (approved):
Good work. I noticed you didn't change much, before we merge, could you please review it again and make sure everything is satisfactory to you? Thanks.

docs-en/chapter_data_structure/number_encoding.md
```diff
@@ -2,11 +2,11 @@
 
 !!! note
 
-    In this book, chapters marked with an * symbol are optional reads. If you are short on time or find them challenging, you may skip these initially and return to them after completing the essential chapters.
+    In this book, chapters marked with an * symbol are optional to read. If you are short on time or find them challenging, you may skip these initially and return to them after completing the essential chapters.
```
K3v123, line 5: Hi YangXuanyi
all of these I'm mentioning are optional:
you can change "In this book, chapters marked with an * symbol are optional to read. If you are short on time or find them challenging, you may skip these initially and return to them after completing the essential chapters."
to ->
"In this book, chapters marked with an asterisk '*' are optional readings. If you have limited time or find them challenging to understand, you may initially skip these chapters and tackle them after completing the essential ones."

docs-en/chapter_data_structure/number_encoding.md
```diff
@@ -2,11 +2,11 @@
...
 
 ## Integer Encoding
 
-In the table from the previous section, we noticed that all integer types can represent one more negative number than positive numbers, such as the `byte` range of $[-128, 127]$. This phenomenon, somewhat counterintuitive, is rooted in the concepts of sign-magnitude, one's complement, and two's complement encoding.
+In the table from the previous section, we noticed that all integer types can represent one more negative number than positive numbers, such as the `byte` range of $[-128, 127]$. This phenomenon seems counterintuitive, and its underlying reason involves knowledge of sign-magnitude, one's complement, and two's complement encoding.
```
K3v123, line 9: "In the table from the previous section, we noticed that all integer types can represent one more negative number than positive numbers, such as the `byte` range of $[-128, 127]$. This phenomenon seems counterintuitive, and its underlying reason involves knowledge of sign-magnitude, one's complement, and two's complement encoding."
can be changed to ->
"In the table presented in the preceding section, we observed that that all integer types can represent one more negative number than positive numbers, for instance, the 'byte' type has a range of [-128, 127]. This phenomenon might seem counterintuitive, and its root cause lies in the knowledge of sign-magnitude, one's complement, and two's complement encoding."

docs-en/chapter_data_structure/number_encoding.md
```diff
@@ -2,11 +2,11 @@
...
 
 Firstly, it's important to note that **numbers are stored in computers using the two's complement form**. Before analyzing why this is the case, let's define these three encoding methods:
```
K3v123, line 11: "Firstly, it's important to note that **numbers are stored in computers using the two's complement form**. Before analyzing why this is the case, let's define these three encoding methods:" can be changed to ->
"Firstly, it's important to note that **numbers are stored in computers using the 'complement' form**. Before delving into the reasons behind this choice, let's define these three encoding methods:"

docs-en/chapter_data_structure/number_encoding.md
```diff
@@ -65,7 +65,7 @@ Adding $1$ to the one's complement of negative zero produces a carry, but with `
 
 One last puzzle is the $[-128, 127]$ range for `byte`, with an additional negative number, $-128$. We observe that for the interval $[-127, +127]$, all integers have corresponding sign-magnitude, one's complement, and two's complement, and these can be converted between each other.
```
K3v123, line 66: "and these can be converted between each other.
Comment" can be changed to ->
"allowing for mutual conversion between them."

docs-en/chapter_data_structure/number_encoding.md
```diff
@@ -65,7 +65,7 @@ Adding $1$ to the one's complement of negative zero produces a carry, but with `
...
 
-However, **the two's complement $1000 \; 0000$ is an exception without a corresponding sign-magnitude**. According to the conversion method, its sign-magnitude would be $0000 \; 0000$, which is a contradiction since this represents zero, and its two's complement should be itself. Computers designate this special two's complement $1000 \; 0000$ as representing $-128$. In fact, the calculation of $(-1) + (-127)$ in two's complement results in $-128$.
+However, **the two's complement $1000 \; 0000$ is an exception without a corresponding sign-magnitude**. According to the conversion method, its sign-magnitude would be $0000 \; 0000$, which is a contradiction because it represents zero, and its two's complement should be itself. Computers designate this special two's complement $1000 \; 0000$ as representing $-128$. In fact, the calculation of $(-1) + (-127)$ in two's complement results in $-128$.
```
K3v123, line 68: "According to the conversion method, its sign-magnitude would be $0000 \; 0000$, which is a contradiction because it represents zero, and its two's complement should be itself." can be changed to ->
"According to the conversion method, its sign-magnitude would be $0000 \; 0000$, indicating zero. This presents a contradiction because its two's complement should represent itself. "

K3v123 (commented):
um, again all these I mentioned are optional.
overall its really accurate to the CN version. (I only went through the parts changed).
thank you for your hard work!

docs-en/chapter_data_structure/number_encoding.md
YangXuanyi: Thank you very much for your suggestion. Among your suggestions, I think this is worth learning from: "with an asterisk '*' are optional readings." I accept your suggestion. The rest is translated as I did before.

docs-en/chapter_data_structure/number_encoding.md
YangXuanyi: Thank you so much. I revisited the meaning and usage of the words “observe” and “notice”. "Observe" is more suitable in this context.

docs-en/chapter_data_structure/number_encoding.md
YangXuanyi: Here I think the original expression is simpler and easier to understand, so I don't want to modify it. But thank you again for providing a new translation idea.

docs-en/chapter_data_structure/number_encoding.md
YangXuanyi: Again, I think the original expression is simpler.

docs-en/chapter_data_structure/number_encoding.md
YangXuanyi: What a great idea. I totally accept that.

YangXuanyi:
Thanks very much for your suggestion. I have given a point by point response as per your suggestion, please check.

YangXuanyi:
> 干得好。我注意到你没有太大变化，在我们合并之前，你能不能再审查一下，确保一切都令你满意？谢谢。

Thanks for the suggestion. I'll review the entire document again to make sure it's correct.

docs-en/chapter_data_structure/number_encoding.md
krahets: @YangXuanyi It’s a tradeoff. We prefer authentic expressions, without introducing complicated words.

Obviously, "allowing for mutual conversion between them." is more authentic.

krahets:
Hi @YangXuanyi, it is not recommended to use force-pushing because it may break the commits' linear history, making it harder to track the changes.

YangXuanyi:
@krahets Sorry. I'll be more careful in the future.

krahets (approved):
Thanks for the PR @YangXuanyi and the review @yuelinxin @K3v123 !

K3v123 (approved):

PR #1062, opened 2024-01-24
fix: Update linkedlist_deque.cpp

Allen-Scai:
There is something error when delete node, should 'delete front' or 'delete rear' outside the body of if cycle.


krahets (approved):
Thanks for the PR!

PR #1064, opened 2024-01-27
Fix the Rust code bug mentioned in discussion 514

night-cruise:
See here 👉 <https://github.com/krahets/hello-algo/discussions/514>


krahets (approved):
Thanks for the fix!

PR #1107, opened 2024-03-04
Fix toc for the webpage of the chapter of computational complexity

nidhoggfgg:
fix the math formula in TOC

before:
屏幕截图 2024-03-04 155313

after:
屏幕截图 2024-03-04 155432

fix both in English and Chinese chapter


krahets:
Thanks for the PR!

I'm seeking a more elegant way to solve the latex rendering issue in the toc. Do you know how to solve this without adding the extra `data-toc-label`?

nidhoggfgg:
this is a problem with python-mardown used by mkdocs, related issues [#849](https://github.com/squidfunk/mkdocs-material/issues/849), [#699](https://github.com/Python-markdown/markdown/issues/699)

except `data-toc-label`, maybe you need to do more work, like writing extensions, ig  
and ASAIK there is no perfect solution yet

here are two more possible solutions (maybe not that good):

1.use javaScript

write these codes to `overrides/javascripts`, tested and they works
```javascript
document$.subscribe(function () {
  document.querySelectorAll('.md-nav__link').forEach(function (link) {
    // only remove the extra charcters when the toc contains sth like O(n) or O(\log n)
    if (link.innerHTML.includes('O(')) {
      link.innerHTML = link.innerHTML.replace(/\\\(/g, '').replace(/\\\)/g, '').replace(/\\/g, '');
    }
  });
});
```

2.programming a extension

lol, remove those `\` or try to render them in nav, it's up to you

btw, in this pr i forgot the space in `平方阶 O(n^2)`, sry for that

krahets (approved):
Thanks for the fix!

PR #1143, opened 2024-03-15
Update insertion_sort.py

tejhande:
Type Annotation Issue: The type annotation list[int] may not be compatible with all Python versions. Use List[int] from the typing module for broader compatibility.

Variable Naming: The variable name base could be more descriptive. Consider renaming it to something like current_value for clarity.

Redundant Comparison: The condition nums[j] > base in the while loop is redundant. It can be simplified to directly compare nums[j] with nums[j+1], improving readability.

These changes improve the code's readability and maintainability without altering its functionality.

### DIFF
--- a/codes/python/chapter_sorting/insertion_sort.py
+++ b/codes/python/chapter_sorting/insertion_sort.py
@@ -1,25 +1,16 @@
-"""
-File: insertion_sort.py
-Created Time: 2022-11-25
-Author: timi (xisunyy@163.com)
-"""
+from typing import List
 
-
-def insertion_sort(nums: list[int]):
-    """插入排序"""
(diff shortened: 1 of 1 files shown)


krahets:
Thanks for the PR!

> Type Annotation Issue: The type annotation list[int] may not be compatible with all Python versions. Use List[int] from the typing module for broader compatibility.

Python>=3.10 is required for this project

> Variable Naming: The variable name base could be more descriptive. Consider renaming it to something like current_value for clarity.

I think `base` is clear enough.

> Redundant Comparison: The condition nums[j] > base in the while loop is redundant. It can be simplified to directly compare nums[j] with nums[j+1], improving readability.

I did not see the changes. Both of the implementations are OK for me.

PR #1157, opened 2024-03-19
Add auto-build workflow for kotlin

curtishd:


.github/workflows/kotlin.yml
```diff
@@ -0,0 +1,18 @@
...
+    branches: [ "main" ]
+    paths: ["codes/kotlin/**/*.kt"]
+  pull_request:
+    branches: [ "main" ]
+    paths: ["codes/kotlin/**/*.kt"]
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
```
krahets, line 13: Could we add windows and macOS as well?

curtishd: Sure

PR #1158, opened 2024-03-19
feat: Add ruby code - chapter "array & linked list"

khoaxuantu:
If this pull request (PR) pertains to **Chinese-to-English translation**, please confirm that you have read the contribution guidelines and complete the checklist below:

- [ ] This PR represents the translation of a single, complete document, or contains only bug fixes.
- [ ] The translation accurately conveys the original meaning and intent of the Chinese version. If deviations exist, I have provided explanatory comments to clarify the reasons.

If this pull request (PR) is associated with **coding or code transpilation**, please attach the relevant console outputs to the PR and complete the following checklist:

arr_linked_list


codes/ruby/chapter_array_and_linkedlist/linked_list.rb
```diff
@@ -0,0 +1,81 @@
...
+File: linked_list.rb
+Created Time: 2024-03-18
+Author: Xuan Khoa Tu Nguyen (ngxktuzkai2000@gmail.com)
+=end
+
+require_relative '../utils/list_node'
+require_relative '../utils/print_util'
+
+# 在链表的节点 n0 之后插入节点 new_node
+def insert(n0, new_node)
```
khoaxuantu, line 11: Here I have to change the variable from `P` to `new_node` because the capitalized letters present constants in Ruby

codes/ruby/chapter_array_and_linkedlist/array.rb
```diff
@@ -0,0 +1,105 @@
+=begin
+File: array.rb
+Created Time: 2024-03-18
+Author: Xuan Khoa Tu Nguyen (ngxktuzkai2000@gmail.com)
+=end
+
+# 随机访问元素
```
krahets, line 7: Is there another comment format to label the headings?

codes/ruby/chapter_array_and_linkedlist/linked_list.rb
krahets: Is that a hard limit?

We'd like to preserve the consistency among multi-languages in terms of variable, function, and class naming.

khoaxuantu: It is a variable mechanic in Ruby. Every variable that begins with a uppercase letter will be treated as a constant in the scope of a class/module. On the other hand, lowercase variables represent local variables, or a method's arguments.

[Ruby's variables and constants](https://www.w3resource.com/ruby/ruby-variables-constants.php)

Defining a method with an uppercase argument violates the rule of Ruby:

Screenshot 2024-03-26 at 08 24 33

So I think we can use `p` to preserve the minimum consistency among multi-languages variable naming

```suggestion
def insert(n0, p)
```

codes/ruby/chapter_array_and_linkedlist/array.rb
krahets: > =begin 随机访问元素 =end

I see. Do you think it is a common usage? And what is the industrial standard style of heading comments for Ruby?

khoaxuantu: > > =begin 随机访问元素 =end
> 
> I see. Do you think it is a common usage? And what is the industrial standard style of heading comments for Ruby?

Actually I barely see this usage in practice. Most of the time when I'm working with Ruby I see people prefer using the second one if they want to write a heading comment

```rb
=begin
随机访问元素
=end
```

It is clear enough to understand that they want to state a heading instead of a normal comment. If it is ok for you, I will push another commit to resolve your 2 reviews now

krahets: I see. How about `### 随机访问元素 ###`?

khoaxuantu: > I see. How about `### 随机访问元素 ###`? A pythonic solution

That's great! We can use it as our own Ruby heading comment convention. With this format, we can even remain compatible with the whole Python sample code.

I would like to briefly summarize the incoming Ruby comment conventions for the project:
- Heading comment: `### Heading ###`
- Inline comment: `#`
- Multiline comment:
```rb
# Abc
# Xyz
# Some thing
```
- File header:
```rb
=begin
File:
Created Time:
Author:
=end
```
- Classes/modules/methods comment (following [Yardoc](https://yardoc.org/guides/index.html) conventions)
```
    # Generate a linked list from an array
    #
    # @param [Array<Integer>] arr
    #
    # @return [ListNode] the root of the linked list
    def generate_from_array(arr) 
    end
```

krahets: Great conclusion!

krahets: By the way, we'd like to avoid using the "Classes/modules/methods comment" to reduce the lines of the code.

codes/ruby/chapter_array_and_linkedlist/linked_list.rb
```diff
@@ -0,0 +1,83 @@
...
+File: linked_list.rb
+Created Time: 2024-03-18
+Author: Xuan Khoa Tu Nguyen (ngxktuzkai2000@gmail.com)
+=end
+
+require_relative '../utils/list_node'
+require_relative '../utils/print_util'
+
+### 在链表的节点 n0 之后插入节点 _p ###
+
```
krahets, line 11: Please remove this empty line for more compact formatting. (Same as others)

codes/ruby/utils/list_node.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def initialize(val=nil, next_node=nil)
+    @val = val || 0
+    @next = next_node
+  end
+end
+
+class LinkedList
+  class << self
+    # Generate a linked list from an array
```
krahets, line 20: Please refer to Python version to use Chinese comments.

(I will update the comments of Java version to Chinese too)

codes/ruby/utils/print_util.rb
```diff
@@ -0,0 +1,21 @@
+=begin
+File: print_util.rb
+Created Time: 2024-03-18
+Author: Xuan Khoa Tu Nguyen (ngxktuzkai2000@gmail.com)
+=end
+
+module Utils
+  class Printer
```
krahets, line 8: Remove the class if it is not necessary.

(In Java, every method should belong to a class)

docs/chapter_array_and_linkedlist/linked_list.md
```diff
@@ -185,6 +185,21 @@
     }
     ```
 
+=== "Ruby"
+
+    ```rb title=""
```
krahets, line 190: Recommend using full heading " ```ruby "

codes/ruby/utils/print_util.rb
khoaxuantu: I see, I wrote them based on the Java files, I will update them based on the Python files instead. 

But I see that the method `linked_list_to_list` in `python/modules/list_node.py` is different from the code in other programming languages, which are `getListNode` instead. Which one should I implement for the Ruby?

krahets: `getListNode()` and `linked_list_to_list()` are both not called in the source code
I'll remove the `getListNode()` for all the languages.

krahets (approved):
Thanks for the clean code! Looking forward to your next PR

docs/chapter_array_and_linkedlist/list.md
```diff
@@ -270,6 +280,15 @@
     nums.items[1] = 0; // 将索引 1 处的元素更新为 0  
     ```
 
+=== "Ruby"
+
+    ```ruby title="list.rb"
+    # 访问元素
+    num = nums[1]
```
krahets, line 287: The comments is not consistent with Python version

docs/chapter_array_and_linkedlist/list.md
```diff
@@ -510,6 +529,26 @@
...
+
+    # 在尾部添加元素
+    nums << 1
+    nums << 3
+    nums << 2
+    nums << 5
+    nums << 4
+
+    # 在中间插入元素
+    nums.insert 3, 6
```
krahets, line 546: Same as above

docs/chapter_array_and_linkedlist/list.md
```diff
@@ -510,6 +529,26 @@
...
+
+    # 删除元素
+    nums.delete_at 3
```
krahets, line 549: Same as above

krahets:
@khoaxuantu I've added Ruby code blocks to the document in #1200 and it's more convenient to add the code.

And I found inconsistency of comments between Ruby and Python. Please fix them and carefully check the code.

Moreover, you don't have to add Ruby code to `en/docs` English version. I'll handle them after completing the Ruby translation.

khoaxuantu:
> And I found inconsistency of comments between Ruby and Python. Please fix them and carefully check the code.
> 
> Moreover, you don't have to add Ruby code to `en/docs` English version. I'll handle them after completing the Ruby translation.

Got it! I have opened a new pull request at #1202 to fix the inconsistent comments. Please help me to review it.

PR #1160, opened 2024-03-22
Modified array.kt

curtishd:


curtishd:
I'm not very good at using the Actions workflow, so can the kotlin.yml config file be submitted a little later?

krahets:
> I'm not very good at using the Actions workflow, so can the kotlin.yml config file be submitted a little later?

OK

krahets:
The code added in this PR should be placed in the document only. I will add kotlin language blocks to the documents so that you can continue the work.

PR #1170, opened 2024-03-24
Add the kotlin blocks to the documents

krahets:


curtishd:
Ok, I will add those after fixing the incorrect comment.

PR #1181, opened 2024-03-26
Update README.md

liruifengv:
Update README

Now supports 13 languages


krahets:
Kotlin is under testing.

PR #1188, opened 2024-03-27
Update top_k.md

skyler-developer:
Top-k问题应是，数组中的第K个最大元素，原文中少了“第”字

The Top-k problem should be that the Kth largest element in the array is missing the word "th" in the original text


krahets:
Not really. In this article, the problem is to find the largest `n` numbers. Please carefully check it.

PR #1203, opened 2024-03-31
fix memory leak in the C code

yang-le:
Compared to the C++ version, I think this C version of 'pop' cannot free the popped-out node correctly when it's the only one remains in the queue. This PR fix this and make it consistent with the C++ version.

### DIFF
--- a/codes/c/chapter_hashing/hash_map_open_addressing.c
+++ b/codes/c/chapter_hashing/hash_map_open_addressing.c
@@ -49,6 +49,9 @@ void delHashMapOpenAddressing(HashMapOpenAddressing *hashMap) {
             free(pair);
         }
     }
+    free(hashMap->buckets);
+    free(hashMap->TOMBSTONE);
+    free(hashMap);
 }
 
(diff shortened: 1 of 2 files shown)


yang-le:
It seems that the hashMap itself was not freed also, commit [e9af854](https://github.com/krahets/hello-algo/pull/1203/commits/e9af854ced9dd46f2be534b58d1b550a816b860b) fix this.

BTW, I think it will be better if we store the TOMBSTONE in a static variable like below instead of mallocing it every time.
``` c
static Pair tombstone = { INT_MAX, NULL };
...
void delHashMapOpenAddressing(HashMapOpenAddressing *hashMap) {
...
if (pair != NULL && pair !=  &tombstone) {
...
````

Gonglja (commented):
Approved @krahets, thanks for the fix.

Gonglja (commented):
Approved @krahets, thanks for the fix.

krahets (approved):
Thanks for the PR @yang-le and the review @Gonglja !

PR #1226, opened 2024-04-05
fix(Rust): no need to clone in hash put fn

rongyi:
in this fn:
`    fn put(&mut self, key: i32, val: String) `

as you can see from the method signature val is owned by this method, so there's no need to clone to put in hash bucket.

and fmt some code to drop unseen space, but it shows in my editor:

图片

### DIFF
--- a/codes/rust/chapter_hashing/hash_map_chaining.rs
+++ b/codes/rust/chapter_hashing/hash_map_chaining.rs
@@ -102,7 +102,6 @@ impl HashMapChaining {
         // 遍历桶，若遇到指定 key ，则更新对应 val 并返回
         for pair in bucket {
             if pair.key == key {
-                // replace value
                 pair.val = val;
                 return;
             }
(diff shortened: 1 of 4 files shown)


codes/rust/chapter_hashing/hash_map_chaining.rs
```diff
@@ -102,17 +102,15 @@ impl HashMapChaining {
         // 遍历桶，若遇到指定 key ，则更新对应 val 并返回
         for pair in bucket {
             if pair.key == key {
-                pair.val = val.clone();
+                // replace value
```
krahets, line 105: Plz do not add the unnecessary comments for the consistency between multiple languages

rongyi: ok

rongyi: updated

krahets (approved):
Thanks for the PR!

PR #1231, opened 2024-04-06
fix: Some Ruby code

khoaxuantu:
- I spotted a wrong Ruby code block in the chapter computational complexity, part **时间复杂度**. It is expect to be `(0...n)` but somehow I wrote `(n...0)` instead.
- I also change a bit of the `remove` method at `array.rb` to be more consistent with other programming languages.


krahets (approved):
Thanks for the fix!

PR #1262, opened 2024-04-10
feat: Add Ruby code - chapter "Searching"

bluebean-cloud:
If this pull request (PR) is associated with **coding or code transpilation**, please attach the relevant console outputs to the PR and complete the following checklist:

- [x] I have thoroughly reviewed the code, focusing on its formatting, comments, indentation, and file headers.
- [x] I have confirmed that the code execution outputs are consistent with those produced by the reference code (Python or Java).
- [x] The code is designed to be compatible on standard operating systems, including Windows, macOS, and Ubuntu.

Ruby_code_searching_run_result

PR #1265, opened 2024-04-10
chore: Update Ruby driver code

khoaxuantu:
In #1262, we have the `binary_search_edge` file that import a method of `binary_search_insertion` file. To prevent the driver code in the `binary_search_insertion` file being triggered when running the `binary_search_edge` file, we should put the driver code into this block:

```rb
### Driver Code ###
if __FILE__ == $0
  # Your code
end
```

I create this PR to update the driver code of all the existing Ruby index files at main.

### DIFF
--- a/codes/ruby/chapter_array_and_linkedlist/array.rb
+++ b/codes/ruby/chapter_array_and_linkedlist/array.rb
@@ -76,32 +76,33 @@ end
 
 
 ### Driver Code ###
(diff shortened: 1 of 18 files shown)


bluebean-cloud (approved):
I think it is no probleam.

krahets (approved):
Thanks for the PR!

PR #1280, opened 2024-04-16
feat(TypeScript): Add the workflow file for CI check

Shyam-Chen:
If this pull request (PR) pertains to **Chinese-to-English translation**, please confirm that you have read the contribution guidelines and complete the checklist below:

- [ ] This PR represents the translation of a single, complete document, or contains only bug fixes.
- [ ] The translation accurately conveys the original meaning and intent of the Chinese version. If deviations exist, I have provided explanatory comments to clarify the reasons.

If this pull request (PR) is associated with **coding or code transpilation**, please attach the relevant console outputs to the PR and complete the following checklist:

- [ ] I have thoroughly reviewed the code, focusing on its formatting, comments, indentation, and file headers.
- [ ] I have confirmed that the code execution outputs are consistent with those produced by the reference code (Python or Java).
- [ ] The code is designed to be compatible on standard operating systems, including Windows, macOS, and Ubuntu.


krahets:
Hi, I added a bug to test the workflow but it still reported success.

codes/typescript/chapter_computational_complexity/space_complexity.ts
```diff
@@ -9,7 +9,7 @@ import { TreeNode } from '../modules/TreeNode';
 import { printTree } from '../modules/PrintUtil';
 
 /* 函数 */
-function constFunc(): number {
+function constFunc(): numer {
```
krahets, line 12: This bug is for testing the workflow

Shyam-Chen:
It seems like executing TypeScript code (`test_all.ts`) might not be necessary anymore, as the code doesn't utilize assertions, so type checking should suffice. Moving the execution to JavaScript could be an option.

codes/typescript/chapter_graph/graph_adjacency_list.ts
```diff
@@ -91,50 +91,47 @@ class GraphAdjList {
     }
 }
 
-// need to add the package @types/node contains type definitions for Node.js, npm i --save-dev @types/node
-if (require.main === module) {
```
krahets, line 95: Why removing this?

codes/typescript/chapter_heap/my_heap.ts
```diff
@@ -122,7 +122,7 @@ class MaxHeap {
 }
 
 /* Driver Code */
-if (require.main === module) {
```
krahets, line 125: Why remove this?

krahets: We added this because other code file import this file

krahets:
Please remove the file changes irrelevant to the CI check to make the semantic of the PR clear.

krahets:
By the way, could you avoid force-push too many times? It's relatively inconvenient to track the history.

Shyam-Chen:
https://docs.deno.com/runtime/manual/runtime/import_meta_api#importmetamain

```ts
require.main === module
// to
import.meta.url === new URL('file://' + process.argv[1]).href
// to
import esMain from 'es-main';
esMain(import.meta)
```

codes/typescript/package-lock.json
```diff
@@ -0,0 +1,144 @@
+{
+    "name": "typescript",
```
krahets, line 2: Although I'm not familiar with TS, it seems too complicated for this project. Are these dependencies necessary for readers?

Shyam-Chen: This is for CI use, users can directly execute the code using `tsx`.

codes/typescript/test_all.ts
```diff
@@ -0,0 +1,57 @@
+import { expandGlob, type WalkEntry } from 'jsr:@std/fs';
```
krahets, line 1: Can this script report compile bugs?

(It seems did not work before.)

codes/typescript/package-lock.json
krahets: There are 9 dependencies in this file. Are they prerequisites? We only need the fundamental language environment In other code languages including JavaScript. What makes Typescript such complex?

Shyam-Chen: In TypeScript, code execution and checking are separate, but VS Code provides built-in checking for the current TypeScript file.

codes/typescript/test_all.ts
krahets: image

Emm, I meant that this script just ran the code without checking it. i.e., `test_all.ts` produced the unexpected behavior.

Shyam-Chen: The code is checked using `tsconfig.json` and executed using the `tsc` command, so the part about executing the code may be removable.

codes/typescript/package-lock.json
krahets: > In TypeScript, code execution and checking are separate, but VS Code provides built-in checking for the current TypeScript file.
Does this reply belong to the other comment?

codes/typescript/test_all.ts
krahets: OK

codes/typescript/tsconfig.json
```diff
@@ -1,8 +1,31 @@
 {
     "compilerOptions": {
```
krahets, line 2: I think the config is too complex. Could we keep some of them default? 

(This book is not for TypeScript professionalism. We want to keep it as simple as possible.)

codes/typescript/package-lock.json
krahets: I'm wondering why we need these dependencies. If it is caused by using ES instead of CommonJS. I'd recommend restoring the previous settings.

codes/typescript/chapter_heap/my_heap.ts
```diff
@@ -4,6 +4,8 @@
  * Author: Justin (xiefahit@gmail.com)
  */
 
+import esMain from 'es-main';
```
krahets, line 7: `import.meta.url.endsWith(process.argv[1])` may be an alternative to avoid requiring `es-main` library.

codes/typescript/package-lock.json
krahets: Can we remove this file from the repo because it may be generated with `npm install`. I think we don't have to lock the dependencies versions for this project strictly.

Shyam-Chen: You can delete or just ignore it, as it's only used for checking the code in CI.

krahets (approved):
Thanks for the PR!

Would you like to add the Javascript's workflow file in another PR?

Shyam-Chen:
> Would you like to add the Javascript's workflow file in another PR?

In JavaScript, only the previously written `test_all.ts` can be used for modification.

PR #1289, opened 2024-04-18
Recommand zed editor for rust

rongyi:


krahets:
Thanks for the suggestion! I think it's enough to learn algorithms using VS Code. And zed editor is not compatible with Windows and Linux.

PR #1291, opened 2024-04-19
Rust: Update btree search

rongyi:


krahets:
The code breaks multi-language consistency.

By the way, the b tree and the binary search tree are two different kinds of trees.

PR #1309, opened 2024-04-28
fix(Rust): Update my_heap.rs

rongyi:
First I find here using pop is better than remove, but I'm hesitate to create a PR for this(in case some inconsistent) until I found cpp is using pop_back() too. so here I am :)


krahets (approved):
Thanks for the PR!

PR #1313, opened 2024-04-30
Fix incorrect method name in permutations_ii.go

xiaomiusa87:
Fix incorrect method name in permutations_ii.go


krahets (approved):
Thanks for the fix!

PR #1329, opened 2024-05-01
Maintain the same semantics as other languages in n_queens.go

xiaomiusa87:
Maintain the same semantics as other languages in n_queens.go


krahets (approved):
Thanks for the fix!

PR #1352, opened 2024-05-09
[Rust] Stick with swap

rongyi:
in select sort we use swap, so we can stick to swap method to swap two value in vector.


krahets (approved):
LGTM

PR #1356, opened 2024-05-10
[Rust] No need to use mut iter

rongyi:
there's no need to use mut iter, just iter is enough

### DIFF
--- a/codes/rust/chapter_sorting/counting_sort.rs
+++ b/codes/rust/chapter_sorting/counting_sort.rs
@@ -10,11 +10,11 @@ include!("../include/include.rs");
 // 简单实现，无法用于排序对象
 fn counting_sort_naive(nums: &mut [i32]) {
     // 1. 统计数组最大元素 m
-    let m = *nums.into_iter().max().unwrap();
+    let m = *nums.iter().max().unwrap();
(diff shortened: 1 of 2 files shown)


codes/rust/chapter_sorting/counting_sort.rs
```diff
@@ -31,16 +31,16 @@ fn counting_sort_naive(nums: &mut [i32]) {
 // 完整实现，可排序对象，并且是稳定排序
 fn counting_sort(nums: &mut [i32]) {
     // 1. 统计数组最大元素 m
-    let m = *nums.into_iter().max().unwrap();
+    let m = *nums.iter().max().unwrap() as usize;
```
krahets, line 34: Why `as usize`?

diucicd: as you can see in the context, m is itended to  be used as usize, in here we cast it to usize instead of multiple case in used place.

rongyi: as you can see in the context, m is itended to  be used as usize, in here we cast it to usize instead of multiple case in used place.

krahets (approved):
LGTM

PR #1363, opened 2024-05-14
[Rust] Normalize mid calculation in case overflow

rongyi:
mid is always calculated with i + (j - i) / 2 in other place, so change it!


krahets:
Could you help also update other languages in this PR?

rongyi:
yes! pleasure! will push to here soon

krahets (approved):
Thanks for the PR!

PR #1411, opened 2024-07-06
Rust: remove unnecessary type conversions

wodray:
There is no need to use `as` to convert the type of ignored value.


krahets (approved):
Thanks for the PR!

PR #1448, opened 2024-07-16
fix: right-left and left-right inversion for the English version of AVL Tree

goerll:
Fix translation mistake of "先左旋後右旋" from "Right-left" to "Left-right" and "先右旋後左旋" from "Left-right" to "Right-left" on the AVL tree page.


krahets (approved):
Thanks for the PR!

PR #1496, opened 2024-08-28
quick_sort pythonic

xiazhicheng:
quick_sort pythonic

- [* ] I have thoroughly reviewed the code, focusing on its formatting, comments, indentation, and file headers.
- [ *] I have confirmed that the code execution outputs are consistent with those produced by the reference code (Python or Java).
- [ *] The code is designed to be compatible on standard operating systems, including Windows, macOS, and Ubuntu.

### DIFF
--- a/codes/python/chapter_sorting/quick_sort.py
+++ b/codes/python/chapter_sorting/quick_sort.py
@@ -111,6 +111,15 @@ class QuickSortTailCall:
                 right = pivot - 1  # 剩余未排序区间为 [left, pivot - 1]
 
 
+def quick_sort(nums: list[int]):
+    """快速排序类（更pythonic）"""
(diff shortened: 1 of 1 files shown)


krahets (commented):
Single-line-solution is cool, but it sacrifices readability.

xiazhicheng:
I believe that for most Python developers, this approach is more readable. What do others think?

krahets:
This book is designed for beginners. However, one-line solutions are relatively difficult for them to understand. We are expected to provide concise step-by-step code which should not contain too much syntactic sugar.

In addition, the pythonic version creates two new lists in each recursive function, resulting in the loss of the in-place feature of quick sort, considered a major advantage of quick sort (being cache-friendly).

PR #1499, opened 2024-09-02
update heap-push-operation description

gopherHsu:
modify the description of heap-push-operation


krahets:
Thanks for the PR!

Here, we focus only on growth trends rather than the absolute heights of the tree. Therefore, I prefer the original expressions.

PR #1609, opened 2025-01-08
fix rust linked list

rongyi:
rust linked_list update

1. in remove, the if let check already exclude the None case, so no need in rust
2. in find, the third argument is only show in rust, the signature should be API level equivalence. hide this in inner function.

happy 2025!


krahets:
Happy new year @rongyi !

krahets (approved):

PR #1625, opened 2025-01-21
Fix python code style issue in linked list deque

wodray:
According to [PEP 8](https://peps.python.org/pep-0008/#programming-recommendations), "Comparisons to singletons like None should always be done with is or is not, never the equality operators."

- [x] This PR represents the translation of a single, complete document, or contains only bug fixes.
- [x] The translation accurately conveys the original meaning and intent of the Chinese version. If deviations exist, I have provided explanatory comments to clarify the reasons.


krahets (approved):
Thanks for the fix!

PR #1665, opened 2025-02-20
Update hash_map_chaining with linked feature

DonglingMo:
This PR enhances the hash_map_chaining implementation by introducing linked insertion order tracking, making it functionally similar to LinkedHashMap. The key changes include:
- Maintaining Insertion Order:
Implemented a doubly linked list to preserve the order of key insertions.
Added prev and next pointers to each entry to ensure efficient traversal.
- Modified put() Method:
When a new key-value pair is added, it is also linked at the end of the list to maintain order.
If an existing key is updated, the insertion order remains unchanged.
- Enhanced remove() Method:
Supports safe removal from both the hash table and the linked list.
Ensures that the insertion order remains valid after deletions.
- Refactored Resizing (extendBuckets())
When resizing occurs, the order of elements is preserved.
The linked list structure is reconstructed after rehashing.

### DIFF
--- a/codes/java/chapter_hashing/hash_map_chaining.java
+++ b/codes/java/chapter_hashing/hash_map_chaining.java
@@ -7,6 +7,7 @@
 package chapter_hashing;
 
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 
 /* 链式地址哈希表 */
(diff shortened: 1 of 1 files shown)


krahets:
Hi @DonglingMo,

In this section, we use arrays to represent the chaining structure for better readability.

PR #1760, opened 2025-05-24
Update Q&A of hash_table and time_complexity Chapter

longsizhuo:
1. Saw a great question in the time complexity section—after answering it, I think it’s worth adding to the Q&A.
2. Also, @Richard-zhao93’s self-asked and self-answered post in the hash map section was excellent.
The suggestion that “if beginners could be introduced to this part earlier, it might make these unfamiliar concepts easier to grasp” is very reasonable.

Thanks for the Reviewing Code, much appreciated!

---

1. 于时间复杂度章节看到一个问题问得很好, 回答后认为可以加入Q&A
2. 于哈希表章节的@Richard-Zhao93 的自问自答非常好, "如果在开头能了解到这部分，对初学者来说，可能更容易接受这些陌生的概念。"的建议很有道理

幸苦大佬CR


krahets (approved):
Thank you, @longsizhuo !

PR #1761, opened 2025-05-27
update rust heap

rongyi:
same as cpp, dont use flag to indicate max heap or min heap, just push value in


krahets (approved):
Thank you for the PR!

PR #1779, opened 2025-07-07
Update array.py

pangpang20:
insert 方法需要先给列表扩展一个位置，否则最后一个元素会丢失


krahets:
The key point here is insertion, we do not have to manage the length of the array.

PR #1807, opened 2025-09-13
fix: String.format() parameter passing error in climbing_stairs_*.java

PXG-XPG:
✅  There are `res` solutions to climb `n` stairs
❎ There are `n` solutions to climb `res` stairs


krahets (approved):
Thank you for the fix!

PR #1812, opened 2025-09-23
docs: add Japanese translate documents

eltociear:
I created Japanese translated documents.
translated from Chinese to Japanese.


krahets (approved):

krahets:
I’ve added your GitHub info to the Japanese landing page to express our gratitude for your contribution. 🥂

PR #1883, opened 2026-04-08
Fix formatting of 'list' in documentation

beatrix-chan:
*Omitted PR template because this is only a formatting issue, not related to translation nor code*

Underline "list" only instead of "A list" in English translation of unit 4.3 under chapter 4 (Array and Linked Lists).


krahets (approved):
Thank you for the fix!

